PR #80, opened 2015-10-17
Changed the bit count from 0 - 9 to 0 - 7 to indicate byte boundaries

michaelbpaulson:
Changed the diagrams from the following to a _better_ format.

current
```
```

Next
```
     0               1               2               3
     0 1 2 3 4 5 6 7 0 1 2 3 4 5 6 7 0 1 2 3 4 5 6 7 0 1 2 3 4 5 6 7
```

If anyone looks at these diagrams the base 10 counting requires an overhead of byte boundaries anyways.  This just streamlines this process.

Here is some motivation for the PR, consider the following example given in Protocol.md

```
    |R|                    Frame Length (optional)                  |
    +-------------------------------+-+-+---------------------------+
    |         Frame Type            |I|M|        Flags              |
    +-------------------------------+-+-+---------------------------+
    |                           Stream ID                           |
                           Depends on Frame Type
```

The `I` and `M` flags in this example require some thought to realize they exist on the 3rd (2) byte as the 0 and 1st bit.  But with the following setup it becomes immediately apparent.

```
     0               1               2               3
     0 1 2 3 4 5 6 7 0 1 2 3 4 5 6 7 0 1 2 3 4 5 6 7 0 1 2 3 4 5 6 7
    |R|                    Frame Length (optional)                  |
    +-------------------------------+-+-+---------------------------+
    |         Frame Type            |I|M|        Flags              |
    +-------------------------------+-+-+---------------------------+
    |                           Stream ID                           |
                           Depends on Frame Type
```

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -47,8 +47,8 @@ transport protocols that provide framing (WebSocket and Aeron), the Frame Length
 For transports that do not provide framing, such as TCP, the Frame Length MUST be included.
 
 ```
-     0                   1                   2                   3
-     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+     0               1               2               3
+     0 1 2 3 4 5 6 7 0 1 2 3 4 5 6 7 0 1 2 3 4 5 6 7 0 1 2 3 4 5 6 7
     +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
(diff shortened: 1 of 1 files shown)


stevegury:
Hi Michael,

While I agree that your change make the reading of byte boundaries clearer, I would prefer sticking to the conventionnal base 10 (used in most protocol RFC). This change would surprized people familliar with protocol ascii-art.

michaelbpaulson:
@stevegury My couple years of professional embedded systems design (I did build an ethernet driver) ascii art was in byte boundary.  But I guess if there is a standard (clearly not designed for those reasoning about / implementing) then there is a standard.

PR #81, opened 2015-10-28
switching the order of the definition for the frame to match the frame order

michaelbpaulson:
While I am implementing some of the js side of things I noticed that some of the definitions appear in different ordering than the frame itself.  This is a simple fix for the ones i have noticed.


stevegury:
Thanks for the fix!

PR #88, opened 2016-04-22
US Typo and reword the frame length requirements

yschimke:
reword the frame length requirements, to allow for protocol with incompatible framing.

For example HTTP/2 has a negotiated SETTINGS_MAX_FRAME_SIZE, so may be running with frames restricted to 16kb. So a transports such as this must fallback to the stream approach.


stevegury:
Hey Yuri, glad to work with you again!
I'm merging the pull request

yschimke:
Thanks, it's a small world. Nice to be downstream of your work again.

Issue #89, opened 2016-05-06
Metadata payload format?


NiteshKant:
#### Motivation

For the reason of flexibility, currently RS protocol does not specify any formats for metadata payload. This is great for flexibility but opens up a way to a wild wild west where everyone in a network call chain can potentially have different metadata formats. Typically, within any infrastructure there would be a need for some standardization of the metadata payload.

This issue intends to gather thoughts around the need for prescribing a metadata payload format.

#### Points to discuss

* Do we need to standardize the metadata payload?
* Do we need to provide a guidance on what is an optimal way of defining metadata payload?
* If we do provide a guidance;
     * What type of data are we looking for? HTTP 1/2 is key value eg.
     * Is [HTTP/2 hpack spec](https://http2.github.io/http2-spec/compression.html#) worthwhile to look at?

#### Possible outcomes

I hope to get to one of the following outcomes of this discussion:

- There is no need to be prescriptive in any way as to what the metadata payload format should be. This is the status quo.
- There is value in standardizing the metadata payload for RS and we come up with a spec. We will come up with a spec and implementation in this case.
- It may be useful to specify a default metadata payload format, which can be overridden if required by mime type negotiation as it exists today. We will come up with a default spec and implementation in this case.

benjchristensen:
The SETUP frame supports negotiation of the MimeType (https://github.com/ReactiveSocket/reactivesocket/blob/master/Protocol.md#setup-frame) for the Metadata. Why is that not sufficient? If a strict format is wanted between two endpoints, then a strict MimeType can be chosen.

NiteshKant:
@benjchristensen it is true that the protocol allows for negotiation. This issue is more about reducing the re-invention that everyone has to do to arrive to an optimal metadata format. If the protocol provides "an option" which is suitable for most of the usecases, there are less things to worry about.

##### Why does this need apply for metadata and not data?

The variance in metadata formats are lesser as compared to data, which depends on the application. In general, metadata are key-value pairs (values can be binary) and it is possible to provide a way that suits most of the usecases. So, it appears to me that there is value in providing a format out of the box.

stevegury:
I think that the protocol is flexible enough to do whatever we want.

That being said, I also agree that, most likely, most of the usage of the metadata will be key/value pairs, so it makes sense to change the default mime-type for the metadata from `UTF-8` to some *very simple* key value encoding, I do not recommend pulling a dependency on hpack.

[...]

PR #90, opened 2016-05-06
Introducing default mime type for metadata.

NiteshKant:
As per discussion in issue #89, this change proposes a default mime-type for metadata payloads.

Existing `Schema.md` is replaced by this new `MimeTypes.md` document which elaborates this new mime type.

This proposal is to specify a default mime type for metadata payload. In absence of a user specifying a mime type (while creating a reactive socket), this default will be used. This does _not_ make the metadata mime type field in the setup frame as optional.


MimeTypes.md
```diff
@@ -0,0 +1,59 @@
+## Mime types
+
+A [Setup Frame](https://github.com/ReactiveSocket/reactivesocket/blob/master/Protocol.md#setup-frame) in the Reactive
+Socket protocol specifies mime types for both [data and metadata](https://github.com/ReactiveSocket/reactivesocket/blob/master/Protocol.md#data-and-metadata) payloads of a frame.
```
stevegury: I think it's better to avoid breaking up the lines explicitly.

MimeTypes.md
```diff
@@ -0,0 +1,59 @@
...
+This document provides a guidance for the mime types of these payloads.
+
+#### Metadata
+
+Reactive Socket provides a default mime type for metadata payloads and all implementations of the protocol MUST use
```
stevegury: ReactiveSocket in one word.

MimeTypes.md
```diff
@@ -0,0 +1,59 @@
...
+##### Overview
+
+The default mime type assumes key-value metadata fields with the following encodings:
+
+* __Key__   : The value is always a UTF-8 string.
+* __Value__ : Value is an opaque binary value and the encoding can vary from one field to another.
+
+##### Name
+
+The name of the default mime type for metadata payloads is __application/x.reactivesocket+kv__. This name MUST be
```
stevegury: Why the "__" before/after?

stevegury:
Other than those 2 comments, LGTM 👍

benjchristensen:
Curious, why define our own encoding format as opposed to adopting an industry standard?

benjchristensen:
> In absence of a user specifying a mime type (while creating a reactive socket), this default will be used. This does not make the metadata mime type field in the setup frame as optional.

Those sentences seem contradictory. How would it not be optional if "in absence of a user specifying" is an option?

NiteshKant:
(Combining responses into a single comment)

@stevegury I will remove the line breaks and use `ReactiveSocket` as a single word. `__` is for making the word bold.

@benjchristensen 
>> Curious, why define our own encoding format as opposed to adopting an industry standard?

The format here is pretty simple with keys as only strings and values as opaque bytes payload. It assumes that if the value is a complex type, it will use some standard format. Most of the standards are pretty heavyweight for this simple requirement. Having said that, if you have something in mind, let me know and I will look at it.

>> Those sentences seem contradictory. How would it not be optional if "in absence of a user specifying" is an option?

I think the language can be improved. Here is what it means:

- The SETUP frame __requires__ a mime type for metadata. This proposal does not relax that requirement i.e. you can not send a SETUP frame without a metadata mime type.
- Any implementation (like the java impl here) MAY provide a way to create a reactive socket without providing a mime type. In such a case, this default mime type should be used.

benjchristensen:
> The format here is pretty simple with keys as only strings and values as opaque bytes payload. It assumes that if the value is a complex type, it will use some standard format. Most of the standards are pretty heavyweight for this simple requirement. Having said that, if you have something in mind, let me know and I will look at it.

My concern is that by reinventing the "simple" approach, are we putting ourselves at risk of repeating what HTTP did with headers in v1 (https://http2.github.io/http2-spec/compression.html#rfc.section.1)?

stevegury:
The difference with HTTP/1 is that with HTTP/1 you __CAN'T__ change the metadata format. Here you still can change the format, we just provide one for free.

NiteshKant:
@benjchristensen I agree it is reinventing a concept that is common for most protocols and HTTP/2 has done a lot of work to "undo" what HTTP/1 did. Lot of HTTP/2's complexity in header compression comes from the fact that it is backward compatible with HTTP/1 and the fact that user agents can keep pretending that they are doing HTTP/1. 
Having said the above, one of the points in the [original PR](https://github.com/ReactiveSocket/reactivesocket/issues/89) was to see whether it is worth plain forklifting HTTP/2 hpack since it is a standard. We can discuss that now if required.

Although, this mime-type is optional, I would like to do the due diligence that it at least works for most cases. If not, there is less value in providing a default that many won't use.

tmontgomery:
Providing a default encoding is probably useful. But I agree with @benjchristensen that it might be best to base it on a standard (or as close as possible). Would a streamlined CBOR or CBOR-compatible encoding work here?

NiteshKant:
@tmontgomery there are only the following data types to be encoded here:

- Map (Implicit as it is a list of K-V pairs)
- UTF-8 string (key)
- Opaque binary payload (value)

I believe the CBOR representation will be pretty close to what is present here. This representation is a bit more compact as we know the types before hand and we just need the lengths. IMHO, picking up CBOR does not provide much value in this case but I am open to it if we have a consensus here that it is better to choose that.

Personally, I believe there is more value in picking up a standard which solves similar problem i.e. representation of headers/metadata.

tmontgomery:
@NiteshKant using the CBOR encoding for those 3 pieces is what I would recommend. It's a standardized (as much as anything) encoding. Can be used with multiple implementations. And has multiple language support already.

Full on CBOR encoding isn't necessary as you point out. However, using a standard encoding that is pretty lightweight, can be referred to, and has multiple implementation support is useful, I think.

By contrast, a specific, new encoding for this makes it a one off that implementations would have to support. Being simple is good, though. But a specific subset of CBOR is simple and a standard. ;)

I would suggest HPACK, but I don't know of any easy way to use a subset of HPACK that is _relatively_ simple.

NiteshKant:
Ok great I am good with CBOR then. 

So, I guess we would have to specify in the spec as to what will be the structure of the metadata payload (map of UTF-8 strings to binary values)? Or we just leave it open and say we use CBOR?

stevegury:
I think using CBOR is fine but I don't like forcing the user to have a dependency to a CBOR library (this is why I proposed a custom implementation).
Maybe this default implementation could be moved to another jar (something like reactivesocket-default-encoding)?

NiteshKant:
@stevegury So, from the java implementation point of view, we have two option:

1. Make [reactivesocket-java](https://github.com/ReactiveSocket/reactivesocket-java) project multi-module and add `reactivesocket-default-encoding` module.
1. Add the `reactivesocket-default-encoding` module to [reactivesocket-java-impl](https://github.com/ReactiveSocket/reactivesocket-java-impl) project.

I would prefer 2 above. Thoughts?

stevegury:
I prefer 2 too.

benjchristensen:
> I don't like forcing the user to have a dependency

Agreed. 

> Add the reactivesocket-default-encoding module to reactivesocket-java-impl project

I prefer this as well if we are providing a default implementation of the default encoding.

Protocol.md
```diff
@@ -571,7 +571,7 @@ The general format for an extension frame is given below.
      0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
     +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
     |R|                    Frame Length (optional)                  |
-    +-------------------------------+-+-+---------------------------+
+    +-+------------------------------+-+-+--------------------------+
```
yschimke: Are the 3 in the middle out of place now?

NiteshKant: Yeah 😢  I will revert this change as the missing `+` is for all frames.

NiteshKant:
So, this is ready for another review. 

The implementation is in this PR:
https://github.com/ReactiveSocket/reactivesocket-java-impl/pull/11

NiteshKant:
If there aren't any objections, I would like to merge this. 
Implementation is merged in the java-impl project.

PR #91, opened 2016-05-13
Metadata Push must be for the connection (stream id 0)

yschimke:
Frame Header Format says stream id must be present, but the spec also says Metadata for a request or response should be in the existing Frame.  So this should be a must.

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -560,7 +560,7 @@ Frame Contents
 
 * __Flags__:
      * (__M__)etadata: Metdadata _always_ present
-* __Stream ID__: May be 0 to pertain to the entire connection.
+* __Stream ID__: Must be 0 to pertain to the entire connection.
 
 ### Extension Frame
 


robertroeser:
I think it can be something other than zero so you can push metadata for a particular stream.

@benjchristensen @stevegury

yschimke:
I thought I was going off the spec

> The following are features of Data and Metadata.
> 
> Metadata can be encoded differently than Data.
> Metadata can be "attached" (i.e. correlated) with the following entities:
> Connection via Metadata Push and Stream ID of 0
> Individual Request or Response

> 
> Metadata Push Frame
> 
> A Metadata Push frame can be used to send asynchronous metadata notifications from a Requester or Responder to its peer. Metadata MUST be scoped to the connection by setting Stream ID to 0.
> 
> Metadata tied to a particular Request, Response, etc. uses the individual frames Metadata flag.

I suggest either removing/rewording the "MUST", or saying that Metadata can be for 3 things instead of 2.  The connection, a stream, or an individual request/response.

stevegury:
👍 
In the early design, metadata was potentially associated with any stream, but later we decided to only allow it at the scope of the session/connection.
(It possible that you'll find area of the spec where it's ambiguous.)

NiteshKant:
Merging since we got two 👍

Issue #59, opened 2015-08-31
GOAWAY for graceful shutdown 


benjchristensen:
Do we need something similar to HTTP/2 GOAWAY?

The use case I have in mind is a server shedding load and disconnecting a connection and letting a client know. 

We don't currently have a connection-level error frame. 

Example use case from HTTP/2 spec:

> Servers are encouraged to maintain open connections for as long as possible but are permitted to terminate idle connections if necessary. When either endpoint chooses to close the transport-layer TCP connection, the terminating endpoint SHOULD first send a GOAWAY (Section 6.8) frame so that both endpoints can reliably determine whether previously sent frames have been processed and gracefully complete or terminate any necessary remaining tasks.

stevegury:
Yes, I just mentioned that in #25, a server needs an explicit way to reject one individual message.
The client has to know that the server didn't process the message, and know that it's safe to retry it (even with non-idempotent messages).

benjchristensen:
The GOAWAY frame though is not for an individual message, it is for the entire connection. 

This is similar to what is being discussed in https://github.com/ReactiveSocket/reactivesocket/issues/50 where we need a connection-level error frame. 

At the individual stream level, we can already use the ERROR frame to reject an individual request.

NiteshKant:
GOAWAY frame is useful for a few things in HTTP/2:

* Stop accepting new streams but finish processing of existing streams.
* Send the last received stream ID before the GOAWAY frame was sent. This makes the client gracefully complete the processing of the existing streams but not send or expect to receive any data for the streams that are initiated after the last received stream by the peer.
* The peer sending the GOAWAY frame need not send any data for the streams initiated post sending GOAWAY frame. This will save sending the ERROR frame for those streams.

If we are not doing these and just indicating error in the frame and close the connection immediately. I think it will have very limited value as opposed to just closing the connection. Graceful shutdown is an important aspect of GOAWAY frame IMO.

tmontgomery:
@NiteshKant has a good point. I'm not sure that graceful shutdown is needed with the LEASE semantic in place. Just send a null LEASE that stops new requests, then send COMPLETEs or ERRORs (for subscriptions).

tmontgomery:
Mentioned in #53 . Repeating here.

We could:

- Define ERROR to be allowed on Stream ID 0.
- Remove SETUP_ERROR as a separate frame type.
- Use ERROR on Stream ID 0 to mean connection level error with connection teardown (?).

[...]

PR #92, opened 2016-06-07
correct stream ids to match HTTP/2

yschimke:
From the HTTP/2 spec

> Streams are identified with an unsigned 31-bit integer.  Streams initiated by a client MUST use odd-numbered stream identifiers; those initiated by the server MUST use even-numbered stream identifiers.  A stream identifier of zero (0x0) is used for connection control messages; the stream identifier of zero cannot be used to establish a new stream.

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -140,7 +140,7 @@ Stream ID value of 0 is reserved for any operation involving the connection.
 A stream ID must be locally unique for a Requester in a connection.
 
 Stream ID generation follows general guidelines for [HTTP/2](https://tools.ietf.org/html/rfc7540) with respect
-to odd/even values. In other words, a client MUST generate even Stream IDs and a server MUST generate odd Stream IDs.
+to odd/even values. In other words, a client MUST generate odd Stream IDs and a server MUST generate even Stream IDs.
 
 ### Frame Types
 


yschimke:
Consider this for discussion.  It seems like something has to change here.  I didn't want to change the code until the spec is agreed.

yschimke:
I'd also like to tighten up the stream id usage in one of two possible ways

1) Match the HTTP/2 Spec - No reuse and monotonically increasing

> The identifier of a newly established stream MUST be numerically
>    greater than all streams that the initiating endpoint has opened or
>    reserved.  This governs streams that are opened using a HEADERS frame
>    and streams that are reserved using PUSH_PROMISE.  An endpoint that
>    receives an unexpected stream identifier MUST respond with a
>    connection error (Section 5.4.1) of type PROTOCOL_ERROR.
> 
>    The first use of a new stream identifier implicitly closes all
>    streams in the "idle" state that might have been initiated by that
>    peer with a lower-valued stream identifier.  For example, if a client
>    sends a HEADERS frame on stream 7 without ever sending a frame on
>    stream 5, then stream 5 transitions to the "closed" state when the
>    first frame for stream 7 is sent or received.
> 
>    Stream identifiers cannot be reused.  Long-lived connections can
>    result in an endpoint exhausting the available range of stream
>    identifiers.  A client that is unable to establish a new stream
>    identifier can establish a new connection for new streams.  A server
>    that is unable to establish a new stream identifier can send a GOAWAY
>    frame so that the client is forced to open a new connection for new
>    streams.

2) Add to the end of Stream Sequences and Lifetimes - "An implementation SHOULD only reuse IDs with prior knowledge of the support by the remote implementation."

stevegury:
I agree, it's better if we match HTTP/2 spec.
The problem is that now we are having a little bit of users, then we need to either:
1. coordinate the changes in our different libraries.
2. use the version field to deal back-compatibility.

I would prefer going for the first solution (because we control the users).
What others think?

NiteshKant:
👍 for monotonically generated, not reused stream IDs (Option 1)

I understand the following needs to be changed to match HTTP/2 spec:
 
- How to handle unexpected stream IDs: Close connection upon receiving a response frame with stream ID that is never generated (more than the last generated stream ID)
- On Stream ID exhaustion: Server MUST close the connection if the stream IDs are exhausted for the server. Client must use a different connection if the stream IDs are exhausted for the client.

Since, we are tightening the usage of Stream IDs, I suggest that we revisit [Handling the Unexpected](https://github.com/ReactiveSocket/reactivesocket/blob/master/Protocol.md#handling-the-unexpected) section. eg: "Receiving a Request frame on a Stream ID that is already in use MUST be ignored." sounds like an error case much like receiving a stream ID that is never generated.

Also, today I do not see in the spec a guidance on how to close a connection i.e. should a ERROR frame (on stream 0) must be sent before closing the connection? I will add another issue for that.

@stevegury 

I agree that incrementing version seems heavy handed for this change. I think we should add a status section in the protocol and we term the current state as "draft 1". Any changes from now can move to the next draft number till we are comfortable that we are good to freeze the protocol.

yschimke:
BTW my thoughts on needing to create new connections after exhausting all stream ids

seuss

stevegury:
@NiteshKant @yschimke I agree with everything.

yschimke:
Sounds like I should leave this with you and you will follow up with issues etc.  Let me know if you would like me to help out or make the changes, but I'll close the PR for now.

yschimke:
@NiteshKant also flipping so 

odd are client generated 
even are server generated

yunong:
Hey all, so sorry to be late to this conversation. 

Restarting the entire connection upon stream id exhaustion seems to be fairly heavy handed. Consider the case where there are long-lived interactions such as Channels or Subscriptions over a connection. These might encapsulate state and be very expensive to restart. Shutting down the connection and starting a new one may not be ideal. What about not shutting down the existing connection?

I wonder if we can also talk about stream_id reuse. I've done some pondering and I _think_ we can't reuse streams_ids because of race conditions and edge cases. Is this provably correct? Not reusing stream ids means we could exhaust the stream IDs fairly quickly. The JS library currently does around 50K RPS, which works out to be just under a day before we exhaust the IDs.

yschimke:
Yeah, I missed that.  I wasn't considering that the server must close the connection.  The connection is still valid to my thinking, just exhausted for new streams.  This would match http/2

> 5.1.1.  Stream Identifiers
> 
> A server
>    that is unable to establish a new stream identifier can send a GOAWAY
>    frame so that the client is forced to open a new connection for new
>    streams.
>    
> 
> 6.8.  GOAWAY
> 
> The GOAWAY frame (type=0x7) is used to initiate shutdown of a
>    connection or to signal serious error conditions.  GOAWAY allows an
>    endpoint to gracefully stop accepting new streams while still
>    finishing processing of previously established streams.  This enables
>    administrative actions, like server maintenance.

NiteshKant:
Just to clarify, the idea was to gracefully shutdown the connection (sending ERROR frame on stream ID 0) as I hinted in [my comment](https://github.com/ReactiveSocket/reactivesocket/pull/92#issuecomment-224638807). I am yet to file an issue of clarifying in the protocol how to do graceful shutdown. 
As per discussion in issue #59 it was determined that ERROR frame on Stream ID 0 and no more leases is equivalent to GOAWAY (graceful shutdown). 

Thanks @yunong for pointing this out.

PR #96, opened 2016-06-10
Fixes #94 (Protocol Status)

### LINKED ISSUE #94: Add a Status for the protocol
##### Problem

ReactiveSocket Protocol spec can not be considered stable as we still see issues like #93 which require minor tweaking or #83 which may require significant change.
If the spec is a moving document, it becomes difficult for people to implement the protocol. We need some kind of versioning of the spec, however, updating major version for every change is pretty heavy handed.

##### Proposal

I suggest the following:

- State in the spec that this is a draft which is currently at version 1 i.e. the spec is a draft-1.
- Create a branch draft-n (which is the next draft for the spec).
- When we see enough changes to the next draft, we merge it to master and create the next draft branch.
- Once we have implementations that are verified in production and we have confidence about stability, we change the status to 1.0.

##### Assumptions

All implementations will have an out-of-band agreement on which draft version are they using between network boundaries. This is not ideal but is better than today where two endpoints have no way of agreeing on which version of the spec they have implemented.

NiteshKant:
As discussed in the issue, this change specifies a status of the protocol as draft.
This also elaborates on the versioning scheme used for the protocol and how to specify the same in the SETUP frame.


stevegury:
👍 This is good.
Can we have a 👍 from our Facebook friends? /cc @yschimke @benjchristensen

Protocol.md
```diff
@@ -206,7 +211,10 @@ Frame Contents
      * (__M__)etadata: Metdadata present
      * (__L__)ease: Will honor LEASE (or not).
      * (__S__)trict: Adhere to strict interpretation of Data and Metadata.
-* __Version__: Version of the protocol.
+* __Version__: Numeric Version of the protocol expressed as two numbers: 
+     * Major Version: 16-bit major version number of the protocol.
+     * Minor Version: 16-bit minor version number of the protocol.
+See [Status](#Staus) for current version.
```
yschimke: typo #Staus

Protocol.md
```diff
@@ -206,7 +211,10 @@ Frame Contents
      * (__M__)etadata: Metdadata present
      * (__L__)ease: Will honor LEASE (or not).
      * (__S__)trict: Adhere to strict interpretation of Data and Metadata.
-* __Version__: Version of the protocol.
+* __Version__: Numeric Version of the protocol expressed as two numbers: 
+     * Major Version: 16-bit major version number of the protocol.
```
yschimke: other sections of the spec use the (16) notation to represent bits.

yschimke:
Something about this feels a bit weird to me, I think mainly having space for 65k major versions.

But I don't have a clear better suggestion that doesn't pick some other arbitrary and more complex logic.  So looks fine to me.  If we get this wrong, we can now bump the major version and change the format of the Setup frame in some backwards incompatible way.

yschimke:
cc @stupaq-fb

stevegury:
Yeah, I initially proposed 4 numbers scheme, but it seems overkill too.

yschimke:
TBH at the point we have 32 major versions of this spec, there will presumably be a Excel spreadsheet somewhere on an NFS share in Netflix with a compatibility matrix

Protocol.md
NiteshKant: So, I didn't want to add another section of versioning scheme and hence used the heading for version to explain it. Anyways, I will modify it.

NiteshKant:
I have addressed the review comments and also added a section on Versioning Scheme and cross version compatibility. PTAL.

stevegury:
What about adding to this paragraph:

> ReactiveSocket assumes that all version changes (major and minor) are backward incompatible.
> A client can pass a version that it supports via the [Setup Frame](#setup-frame)
> It is up to a server to accept clients of lower versions than what it supports.

The main purpose of spliting the version number in two parts, is to iterate over the design of a new major version by quickly incrementing minor version. Every released/stable version of ReactiveSocket should have a minor version equal to zero.

benjchristensen:
cc @tmontgomery for review and input.

NiteshKant:
@stevegury 

>> The main purpose of spliting the version number in two parts, is to iterate over the design of a new major version by quickly incrementing minor version.

I feel this is like repeating the obvious but if it makes sense I don't have a strong preference for either.

>> Every released/stable version of ReactiveSocket should have a minor version equal to zero.

You mean till we move to 1.0?

I feel keeping the statement simple that all versions are backward incompatible is more effective than trying to break down as to what major and minor means as there really isn't any difference in semantics.

stevegury:
Ok, simpler is better.

NiteshKant:
@benjchristensen, @stevegury Since we didn't hear anything from Todd, should we go ahead with this?

benjchristensen:
Todd and I are meeting tomorrow, so I'd like to wait for him to have a chance to review. Last week his was at QCon NY so likely didn't have a chance to review anything.

tmontgomery:
The general idea here is good. But if we split the version field, we should explain what major and minor mean in the spec and what the assumed behavior of a major (and minor) bump means. These can and should be general guidelines, not hard and fast rules.

The version field(s) will be used beyond our iteration and evolution on the protocol, though. So, what value does the field(s) bring after the protocol has become stable in use?

NiteshKant:
Thanks @tmontgomery for the review and welcome back 😄 

>> But if we split the version field, we should explain what major and minor mean in the spec and what the assumed behavior of a major (and minor) bump means. 

This is what I added in this change:

_ReactiveSocket assumes that all version changes (major and minor) are backward incompatible. A client can pass a version that it supports via the Setup Frame It is up to a server to accept clients of lower versions than what it supports._

@stevegury suggested additional information in [his comment](https://github.com/ReactiveSocket/reactivesocket/pull/96#issuecomment-225339131) but my opinion was that keeping it simpler and open is better, but if you have a suggestion on wording, I am open to that.

>> So, what value does the field(s) bring after the protocol has become stable in use?

So there are two conflicting goals here:

1. Have the ability to iterate on the protocol (1.x and thereof). However, we do not want to land on version `238` when it is stable :)  
2. Protocol implementors should have a better idea when things change (These changes can be behavior change or clarification in wording.) and should be able to reject anything that do not understand.

Since these goals are conflicting with a single version field, this change introduced 2 fields (full semantic versioning of 4 fields seems an overkill). 

We can put this reasoning in the spec or have people refer to this issue if at all they want to know why there are 2 fields. I am open to both.

PR #112, opened 2016-08-19
Sync draft-2 branch to master

NiteshKant:


NiteshKant:
Since, this is a simple sync from master, I am just merging it in.

NiteshKant:
Duh, we can only squash and commit here.

NiteshKant:
@benjchristensen So, should we rebase and push directly to sync branches? I surely don't want to squash commits from master.

benjchristensen:
I don't understand the context of what's going on here :-) This should definitely be redone/rebased/whatever so we don't have all those duplicate commits.

benjchristensen:
Oh, you're going TO draft-2. Why? What's the reason for the other branch?

NiteshKant:
@benjchristensen yeah should have added more context.

So, I mentioned in #94 the workflow for changes to the protocol. 
As we do not publish any artifacts and the Protocol spec is the artifact, it will be good if the changes to the spec are batched. It will be easier for the implementors to implement a version and users to refer to a "stable" version of the spec.

I suggested that we make changes to the protocol in a different branch and then promote that to master on a schedule (github milestone). This will make it easy to track changes.

The workflow is very similar to what HTTP/2 did.

I think anyways the name of the branch is wrong, we decided to use `major.minor` format as opposed to `draft-n`, so I will delete this branch and create a fresh one from master.

Issue #111, opened 2016-08-15
Eliminate request/subscription?


benjchristensen:
I have found that the distinction between `request/stream` and `request/subscription` confuses people. This has been discussed somewhat in the past (https://github.com/ReactiveSocket/reactivesocket/issues/40), but the original reason for differentiating between `request/stream` and `request/subscription` was primarily:

- clarify of intent in the client API of a finite stream versus infinite subscription
- enforcement on server of a subscription not emitting `onComplete`

There are not however any performance benefits to the distinction as there are for `request/response` vs `fire-and-forget` vs `request/stream` vs `request/channel`.

Since I end up having to constantly explain why `request/stream` and `request/subscription` exist, I think it is achieving the opposite of the intended goal and we should just eliminate `request/subscription`. The `request/stream` semantics can be used for both finite and infinite use cases and thus can serve subscriptions just by not emitting an `onComplete`.

If we drop it we can then clearly say that ReactiveSocket supports 4 interaction models without explaining the duplication of 2 of 5:

- request/response
- fire-and-forget
- request/stream
- channel

Does anyone else think it is worth simplifying the protocol and API by eliminating `request/subscription`?

NiteshKant:
Strong 👍  for this 😄

yschimke:
+1 

![image](https://cloud.githubusercontent.com/assets/231923/17690246/35261082-6385-11e6-99db-aa01a5dcac22.png)

tmontgomery:
👍 I think this does make things cleaner.

benjchristensen:
Okay, then should I submit a PR for the spec to remove `request/subscription`?

lehecka:
I personally liked the differentiation between stream and subscription, but I don't have enough practical experience to fight for it. I am fine if we simplify the API and remove the extra code.

benjchristensen:
@stevegury @robertroeser your thoughts?

nickmahilani:
👍 for eliminating request/subscription. I think this will simplify the protocol.

drewhk:
+1 from my side as well. Simplicity is desirable at the protocol level.

benjchristensen:
Pretty strong support for this, so I'll eventually submit a PR with the changes to the spec ... unless someone wants to tackle it sooner than I can get to it. If so let me know you're doing so.

NiteshKant:
@benjchristensen I can take this up but let me merge the PR #96 to introduce the protocol status so that we can do this on the next version.

[...]

PR #113, opened 2016-08-22
Fixes Issue #111

NiteshKant:
As discussed in #111, this change removes request-subscription interaction model from the protocol.

**Frame type values have changed.** I did not want to leave a hole in the frame type flags, so I changed the flags specified after `request-subscription`. Yays/Nays?

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -111,17 +111,13 @@ Modern distributed system topologies tend to have multiple levels of request fan
 
 Let's imagine a server responsible for computing the nth digit of Pi. A client sends a request to that server but realizes later that it doesn't want/need the response anymore (for arbitrary reasons). Rather than letting the server do the computation in vain, it can cancel it (the server may not even have started the work).
 
-#### What are example use cases of topic subscription (and push notification)?
-
-Let's imagine a chat server, you want to receive all the messages said in the chat server but you don't want to poll or continuously poll (long polling technique). Subscription is the perfect use case for that.
-
 #### What are example use cases of fire-and-forget versus request-response?
 
 Some requests don't require a response, and when it's fine to simply ignore any failure to send them, fire-and-forget is the right solution. An example could be non-critical metrics tracking in environments where UDP is inappropriate.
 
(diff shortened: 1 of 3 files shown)


yschimke:
I'm Yay, we agreed that each new draft of the protocol spec could break compatibility.  And servers that want to support two versions can make the Frame Type logic conditional on the version number in the setup frame.

I know this will likely cause some issue, but this is the chance to fix things like this as we are pre 1.0

NiteshKant:
Great! Thanks @yschimke !

@benjchristensen @stevegury @robertroeser @tmontgomery WDYT?

stevegury:
👍

FAQ.md
```diff
@@ -111,17 +111,13 @@ Modern distributed system topologies tend to have multiple levels of request fan
 
 Let's imagine a server responsible for computing the nth digit of Pi. A client sends a request to that server but realizes later that it doesn't want/need the response anymore (for arbitrary reasons). Rather than letting the server do the computation in vain, it can cancel it (the server may not even have started the work).
 
-#### What are example use cases of topic subscription (and push notification)?
```
benjchristensen: This use case is still applicable, it should just be solved with request/stream instead of request/subscription.

Motivations.md
```diff
@@ -116,34 +116,24 @@ Usage can be thought of like this:
...
 
 Extending from request/response is request/stream, which allows multiple values to be streamed back. Think of this as a "collection" or "list" response, but instead of getting back all the data as a single response, each element is streamed back in order.
+A stream may be finite or infinite. An infinite stream will not have any terminal event but may cancel.
 
 Use cases could include things like:
 
 - fetching a list of videos
 - fetching products in a catalog
 - retrieving a file line-by-line
+- Infinite event streams that continuously emit, such as metrics, or stock prices.
```
benjchristensen: nit: the other bullet points start with lower-case, this is capitalized

benjchristensen:
👍

benjchristensen:
Thanks for making these changes.

PR #114, opened 2016-09-04
update Akka Streams link

patriknw:


drewhk:
Ah, thanks, 1.0-M4 is ancient.

PR #124, opened 2016-11-14
Add additional spec for 'request N' format

martinsix:
Currently, the format of the request N is ambiguous, and the allowed ranges are not specified. This causes problems with cross platform streams and subscriptions.

### DIFF
--- a/Motivations.md
+++ b/Motivations.md
@@ -43,7 +43,7 @@ This protocol is designed for use both in datacenter, server-to-server, use case
 
 ##### "Reactive Stream" `request(n)` Async Pull
 
-This first form of flow control is suited to both server-to-server and server-to-device use cases. It is inspired by the Reactive Streams [Subscription.request(n)](https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.0/README.md#3-subscription-code) behavior. [RxJava](https://github.com/ReactiveX/RxJava/wiki/Backpressure#how-a-subscriber-establishes-reactive-pull-backpressure), [Reactor](https://github.com/reactor/reactor), and [Akka Streams](http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0-M2/stream-design.html) are examples of implementations using this form of "async pull-push" flow control.
+This first form of flow control is suited to both server-to-server and server-to-device use cases. It is inspired by the Reactive Streams [Subscription.request(n)](https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.0/README.md#3-subscription-code) behavior. [RxJava](https://github.com/ReactiveX/RxJava/wiki/Backpressure#how-a-subscriber-establishes-reactive-pull-backpressure), [Reactor](https://github.com/reactor/reactor), and [Akka Streams](http://doc.akka.io/docs/akka/2.4/scala/stream/index.html) are examples of implementations using this form of "async pull-push" flow control.
 
 ReactiveSocket allows for the `request(n)` signal to be composed over network boundaries from requester to responder (typically client to server). This controls the flow of emission from responder to requestor using Reactive Stream semantics at the application level and enables use of bounded buffers so rate of flow adjusts to application consumption and not rely solely on transport and network buffering.
 
--- a/Protocol.md
+++ b/Protocol.md
@@ -82,7 +82,7 @@ is not present if the Frame Length is not used [see below](#frame-length).
 reception. Flags generally depend on Frame Type, but all frame types must provide space for the following flags:
      * (__I__)gnore: Ignore frame if not understood
      * (__M__)etadata: Metadata present
-* __Stream ID__: (32) Stream Identifier for this frame or 0 to indicate the entire connection.
+* __Stream ID__: (32) Positive signed integer representing the stream Identifier for this frame or 0 to indicate the entire connection.
 
 __NOTE__: Byte ordering is assumed to be big endian.
 
@@ -141,7 +141,7 @@ If Metadata Length is greater than this value, the entire frame MUST be ignored.
     +---------------------------------------------------------------+
 ```
 
-* __Metadata Length__: (31 = max 2,147,483,647 bytes) Length of Metadata in bytes. Including Metadata header. The __R__ bit is reserved and must be set to 0.
+* __Metadata Length__: (31 = max 2,147,483,647 bytes) unsigned positive Integer representing the length of Metadata in bytes. Including Metadata header. The __R__ bit is reserved and must be set to 0.
 
 ### Stream Identifiers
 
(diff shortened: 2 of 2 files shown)


benjchristensen:
Changes look good, and are needed. I would also like the spec to explain the difference of using 2^31-1 vs 2^63-1 in ReactiveStreams, and that this means it doesn't work with the magic number for "infinite streams" that the ReactiveStreams spec has. 

Also, please make the title of this PR represent what the change is.

benjchristensen:
This section should be referenced: https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.0/README.md#3-subscription-code

We should explain that 2^31-1 is being used here so that we only need 4 bytes instead of 8 when doing network transfer, and that it is determined that byte size and flow control is deemed more important over the network than supporting 2^63-1 and the "magic number".

tmontgomery:
A small paragraph explaining how the field maps to ReactiveStreams sounds appropriate.

yschimke:
Should you clarify TTL and Stream ID as well?

benjchristensen:
> Should you clarify TTL and Stream ID as well?

All numbers should be clearly identified as signed since this is polyglot and needs to work between languages that don't support unsigned.

benjchristensen:
@tmontgomery Is there any reason to try and map something like 2^31-1 to the ReactiveStreams 2^63-1 magic number? I don't think the confusion is worth it. The perf benefit only really applies to limited in-memory scenarios. It seems those reasons go away when going over the network with the additional overhead involved.

benjchristensen:
@tmontgomery Also, I'm not super thrilled by this wording in ReactiveStreams:

> A demand equal or greater than 2^63-1 (java.lang.Long.MAX_VALUE) MAY be considered by the Publisher as “effectively unbounded”[3].

The "or greater than" implies either unsigned (impossible in Java) or accumulated. I assume (but can't remember specifically) that we landed on that language as once the number is that big it doesn't really matter, hence the "effectively unbounded". 

Reason I bring this up ... in ReactiveSocket, if we accumulate to higher than 2^31-1, should that be an error or be considered "effectively unbounded". I prefer error since we are only at 2^31-1, and not 2^63-1.

tmontgomery:
Yes, all fields should be explained and unambiguous.

@benjchristensen yes. Agreed. Sorry, I misspoke. I meant that the definition of the `N` field should be expanded to be clear wrt how it is defined relative to the ReactiveStreams definition.

tmontgomery:
@benjchristensen I would feel more comfortable with error. Once `N` reaches 2^31-1 it is capped out as a value. And 2B is a pretty big window....

benjchristensen:
Agreed. Let's error if we exceed 2^31-1

tmontgomery:
If we need `INFINITE`, we can always define it to be 2^31-2 or similar. i.e. we pick a value to represent it.

benjchristensen:
I prefer not having an obscure magic number like that. Since code has to support 'n' anyways is there a legit reason here to signal infinite? I don't think the perf reasons apply here as they can in synchronous use of ReactiveStreams.

NiteshKant:
@martinsix can you target this PR to branch [0.2](https://github.com/ReactiveSocket/reactivesocket/tree/0.2)? That is the next version of the spec. May be we break the PR into two:

1) Clarify language (signed/unsigned positive numbers). This isn't backward incompatible.
2) Error on reaching `2^31 - 1` which is a backward incompatible change.

__Re: Error on `2^31-1`__

As it stands today, in absence of a flow control operator (like buffer eg), a plain subscribe in RxJava will send a `Long.MAX_VALUE` for `requestN` which after this change will start emitting an error. Although, I understand the reasoning for not defining a concept of "unbounded" but it looks to me that emitting an error will create an unnecessary burden on the application to make sure that there isn't a path in which `Long.MAX_VALUE` is requested. 
IMO, at the protocol level, we should just consider `2^31 - 1` as another number and not have any specific behavior associated (infinity or error) with it. Perf optimizations of turning off flow control don't apply also because transport isn't unbounded anyways and an implementation has to handle that flow control in any case. 
Needless to say that an implementation should guard against overflows.

benjchristensen:
> can you target this PR to branch 0.2? That is the next version of the spec.

Curious, why are we using a branch and not just working on master for this? The spec has not been locked down yet, and there is no "release" process that affects running code.

benjchristensen:
> 2) Error on reaching 2^31 - 1 which is a backward incompatible change.

If it's breaking that's only because it was previously unspecified and thus a choice of each implementation. Correct?

benjchristensen:
> As it stands today, in absence of a flow control operator (like buffer eg), a plain subscribe in RxJava will send a Long.MAX_VALUE for requestN which after this change will start emitting an error. Although, I understand the reasoning for not defining a concept of "unbounded" but it looks to me that emitting an error will create an unnecessary burden on the application to make sure that there isn't a path in which Long.MAX_VALUE is requested. 

RxJava v1 Long.MAX_VALUE doesn't work anyways since that is 2^63-1 and ReactiveSocket only supports 2^31-1. So ReactiveSocket impls always have to manage the bridging from ReactiveStreams impls. 

>  it looks to me that emitting an error will create an unnecessary burden on the application to make sure that there isn't a path in which Long.MAX_VALUE is requested. 

Thus, that is already the case. The ReactiveSocket implementation has to deal with this, not the application code.

> IMO, at the protocol level, we should just consider 2^31 - 1 as another number 

Agreed that it should not signal "infinity". But we can't let overflow happen, thus an error must error if it overflows, otherwise flow control is broken as it can no longer be reasoned about. As you say:

> Needless to say that an implementation should guard against overflows.

Thus, the implementation should emit `onError` if `request(n)` is either larger, or accumulates larger than 2^31-1.

NiteshKant:
>>> Curious, why are we using a branch and not just working on master for this? The spec has not been locked down yet, and there is no "release" process that affects running code.

(We had discussed this in #94)

Agreed the protocol is not locked down yet but there are implementations which are not always up-to-date with the protocol spec. eg: java implementation have not yet removed request-subscription but it is removed from the protocol (0.2). It is useful to have a spec that correctly represents the implementation and hence the branching and version strategy.

NiteshKant:
>>> If it's breaking that's only because it was previously unspecified and thus a choice of each implementation. Correct?

Yes that is correct the behavior was unspecified before and now it is specified to be stricter (error).

>>> Thus, the implementation should emit onError if request(n) is either larger, or accumulates larger than 2^31-1.

I overlooked the accumulation case (duhh). I agree that accumulation to larger than `2^31-1` must emit error.

NiteshKant:
@martinsix can you rebase the changes to 0.2?

NiteshKant:
@tmontgomery do you want to pull in this as part of your changes too?

tmontgomery:
Yes, these should be brought in.

benjchristensen:
Based on the approval of @NiteshKant and @tmontgomery, I have resolved the conflicts and will merge.

PR #127, opened 2016-12-16
Changes for layout and proposed 1.0 changes.

tmontgomery:
- Pull out framing into own section. 
- Move StreamID to start RS Header.
- Reduce size of frame length, frame type, flags, and metadata length.
- Remove REQUEST_SUB. 
- Add resume fields to SETUP and KEEPALIVE.

NOTE: Additional frame types for RESUME and RESUME_OK will be in separate PR that details the rest of resuming operation. The fields and frame types are mentioned here for completeness when resume is not used.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -181,9 +181,9 @@ When respecting the lease semantics, establishing a ReactiveSocket between a cli
 You may want to pass data to your application at ReactiveSocket establishment, rather than reimplementing a connect protocol on top of ReactiveSocket, ReactiveSocket provides you the possibility to send information alongside the SETUP frame.
 For instance, this can be used by a client to send its credentials.
 
-#### Why those 5 interaction models?
+#### Why those multiple interaction models?
 
-Those 5 interaction models could be reduced to just one "request-channel". Every other interaction model is a subtype of request-channel, but they have been specialized for two reasons:
+The interaction models could be reduced to just one "request-channel". Every other interaction model is a subtype of request-channel, but they have been specialized for two reasons:
 
 - Ease of use from the client point of view.
 - Performance.
--- a/FAQ.md
+++ b/FAQ.md
@@ -111,17 +111,13 @@ Modern distributed system topologies tend to have multiple levels of request fan
 
 Let's imagine a server responsible for computing the nth digit of Pi. A client sends a request to that server but realizes later that it doesn't want/need the response anymore (for arbitrary reasons). Rather than letting the server do the computation in vain, it can cancel it (the server may not even have started the work).
 
-#### What are example use cases of topic subscription (and push notification)?
+#### What are example use cases of request-stream?
 
-Let's imagine a chat server, you want to receive all the messages said in the chat server but you don't want to poll or continuously poll (long polling technique). Subscription is the perfect use case for that.
+Let's imagine a chat server, you want to receive all the messages said in the chat server but you don't want to poll or continuously poll (long polling technique). Another example might be we want to listen to a particular chat room and ignore all other messages.
 
 #### What are example use cases of fire-and-forget versus request-response?
 
 Some requests don't require a response, and when it's fine to simply ignore any failure to send them, fire-and-forget is the right solution. An example could be non-critical metrics tracking in environments where UDP is inappropriate.
-
-#### What are example use cases of request-stream?
-
-Let's use the same example as for subscription, the chat server, but this time we want to subscribe to a particular chat room and ignore all other messages.
 
 #### Why Binary?
 
--- a/FAQ.md
+++ b/FAQ.md
@@ -181,7 +181,7 @@ When respecting the lease semantics, establishing a ReactiveSocket between a cli
 You may want to pass data to your application at ReactiveSocket establishment, rather than reimplementing a connect protocol on top of ReactiveSocket, ReactiveSocket provides you the possibility to send information alongside the SETUP frame.
 For instance, this can be used by a client to send its credentials.
 
-#### Why those multiple interaction models?
+#### Why multiple interaction models?
 
 The interaction models could be reduced to just one "request-channel". Every other interaction model is a subtype of request-channel, but they have been specialized for two reasons:
 
--- a/Protocol.md
+++ b/Protocol.md
@@ -1,7 +1,7 @@
 ## Status
 
 This protocol is currently a draft for the final specifications. 
-Current version of the protocol is __1.0__ (Major Version: 1, Minor Version: 0).
+Current version of the protocol is __0.2__ (Major Version: 0, Minor Version: 2).
 
 ## Introduction
 
@@ -15,7 +15,7 @@ ReactiveSockets assumes an operating paradigm. These assumptions are:
 
 Key words used by this document conform to the meanings in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
-Byte ordering is assumed to be big endian.
+Byte ordering is big endian for all fields.
 
 ## Terminology
 
@@ -55,7 +55,7 @@ The following are features of Data and Metadata.
     - Connection via Metadata Push and Stream ID of 0
     - Individual Request or Response
 
-## Framing Protocol
+## Framing
 
 ### Transport Protocol
 
@@ -70,12 +70,11 @@ have no further frames sent and all frames will be ignored.
 
 ReactiveSocket as specified here only allows for TCP, WebSocket, and Aeron as transport protocols.
 
-### Framing Protocol
+### Framing Protocol Usage
 
-Some of the supported transport protocols for ReactiveSocket may not support specific framing of messages. For these protocols, a framing protocol must be
-used with the ReactiveSocket frame.
+Some of the supported transport protocols for ReactiveSocket may not support specific framing that preserves message boundaries. For these protocols, a framing protocol must be used with the ReactiveSocket frame that prepends the ReactiveSocket Frame Length.
 
-The framing header MUST be omitted if the transport protocol preserves message boundaries e.g. provides compatible framing. If, however, the transport protocol only provides a stream abstraction or can merge messages without preserving boundaries, or multiple transport protocols may be used, then the frame header MUST be used.
+The frame length field MUST be omitted if the transport protocol preserves message boundaries e.g. provides compatible framing. If, however, the transport protocol only provides a stream abstraction or can merge messages without preserving boundaries, or multiple transport protocols may be used, then the frame header MUST be used.
 
 |  Transport Protocol            | Frame Length Field Required |
 |:-------------------------------|:----------------------------|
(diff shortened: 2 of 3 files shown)


tmontgomery:
Please check for typos and errors specifically. Lots of changes.

NiteshKant:
@tmontgomery I will create a tag for 0.1 from master as it stands today. 0.2 branch has the change to remove request-subscription, should I pull it into master (after tagging)?

tmontgomery:
@NiteshKant we could. Or I just removed it anyway. Maybe close the old PR and go from this.

benjchristensen:
Scanned through, this will take a more focused sit-down review which I'll do as soon as I can. 

Thank you very much @tmontgomery for submitting this.

NiteshKant:
@tmontgomery there are a few references to request-subscription  in `FAQ.md` and `Motivation.md` in the [other branch](https://github.com/ReactiveSocket/reactivesocket/tree/0.2) but yeah I guess you can include them here too. I will leave it to you.

tmontgomery:
@NiteshKant ah. I didn't edit anything but the spec.

Protocol.md
```diff
@@ -99,17 +70,73 @@ have no further frames sent and all frames will be ignored.
...
 
-The presence of the Frame Length field (and corresponding __R__ bit) is determined by the transport protocol being used. The frame length field MUST be omitted if the transport protocol preserves message boundaries e.g. provides compatible framing. If, however, the transport protocol only provides a stream abstraction or can merge messages without preserving boundaries, or multiple transport protocols may be used, then the frame length field MUST be used.
+The framing header MUST be omitted if the transport protocol preserves message boundaries e.g. provides compatible framing. If, however, the transport protocol only provides a stream abstraction or can merge messages without preserving boundaries, or multiple transport protocols may be used, then the frame header MUST be used.
 
 |  Transport Protocol            | Frame Length Field Required |
 |:-------------------------------|:----------------------------|
 | TCP                            | __YES__ |
 | WebSocket                      | __NO__  |
 | Aeron                          | __NO__  |
+| HTTP/2                         | __NO__  |
```
yschimke: Isn't this awkward in practice because of the common presence of HTTP/2 proxies.  The kind of thing that will work great until it suddenly doesn't?

Protocol.md
```diff
@@ -99,17 +70,73 @@ have no further frames sent and all frames will be ignored.
...
+
+When using a transport protocol providing framing, the ReactiveSocket frame is simply encapsulated into the transport protocol messages directly.
+
+```
+    +-----------------------------+-----------------------
+    | ReactiveSocket Frame Header | Depends on Frame Type
+    +-----------------------------+-----------------------
+```
+
+When using a transport protocol that does not provide compatible framing, the Frame Length must be pre-pended to the ReactiveSocket Frame.
```
yschimke: is pre-pended a hyph-enated word?

Protocol.md
tmontgomery: This is from a pure framing perspective. Lots more detail of how to run over HTTP/2 would be needed. But framing is not one of them.

Protocol.md
tmontgomery: hmmm. That is a good question. Let me check into it.

Protocol.md
```diff
@@ -350,18 +382,21 @@ Frame Contents
...
+    +                                                               +
+    |                                                               |
     +---------------------------------------------------------------+
                                   Data
 ```
 
 * __Flags__:
      * (__M__)etadata: Metadata __never__ present
+* __Last Received Position__: (64) Resume Last Received Position (optional. Set to all 0s when not supported.)
```
tmontgomery: Same comment as the token wrt optional (omitted) fields.

Protocol.md
```diff
@@ -15,6 +15,8 @@ ReactiveSockets assumes an operating paradigm. These assumptions are:
 
 Key words used by this document conform to the meanings in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
+Byte ordering is assumed to be big endian.
```
tmontgomery: No. But this is important to mention for clarity.

Protocol.md
```diff
@@ -198,17 +226,23 @@ Frame Contents
      0                   1                   2                   3
      0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
     +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-    |R|                    Frame Length (optional)                  |
-    +-------------------------------+-+-+-+-+-----------------------+
-    |     Frame Type = SETUP        |0|M|L|S|       Flags           |
-    +-------------------------------+-+-+-+-+-----------------------+
-    |                          Stream ID = 0                        |
-    +-------------------------------+-------------------------------+
+    |                           Stream ID                           |
```
tmontgomery: hmmm. Let me think on this one. Good point.

Protocol.md
```diff
@@ -133,15 +160,15 @@ If Metadata Length is greater than this value, the entire frame MUST be ignored.
      0                   1                   2                   3
      0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
     +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-    |R|                       Metadata Length                       |
+    |              Metadata Length                  |
     +-+-------------------------------------------------------------+
```
NiteshKant: format: Reduce width of other sections below too?

Protocol.md
NiteshKant: That's a pretty interesting point. The positive is certainly that it will reduce the data transferred for leases, keep-alive but the negative would be inconsistency and bugs related to missing stream ID counted as stream 0. May be hard to debug these bugs.

FAQ.md
```diff
@@ -185,9 +181,9 @@ When respecting the lease semantics, establishing a ReactiveSocket between a cli
 You may want to pass data to your application at ReactiveSocket establishment, rather than reimplementing a connect protocol on top of ReactiveSocket, ReactiveSocket provides you the possibility to send information alongside the SETUP frame.
 For instance, this can be used by a client to send its credentials.
 
-#### Why those 5 interaction models?
+#### Why those multiple interaction models?
```
NiteshKant: nit: Remove "those"?

Protocol.md
```diff
@@ -1,7 +1,7 @@
 ## Status
 
 This protocol is currently a draft for the final specifications. 
-Current version of the protocol is __0.1__ (Major Version: 0, Minor Version: 1).
+Current version of the protocol is __1.0__ (Major Version: 1, Minor Version: 0).
```
NiteshKant: This is something I don't know how to handle i.e. We are in the __process__ of going to `1.0` but this spec is not yet `1.0`. Anyone coming and reading this spec would assume that it is `1.0` as the version says so. 
This is the reason I thought having a current version branch and WIP branch is useful.

Protocol.md
robertroeser: Good point Nitesh. Zero is even so we could just say 'stream 0' is where the start of the even frames ids. Rename 'stream 0' as admin stream or something along those lines.

Protocol.md
benjchristensen: I'm okay with this remaining something like 0.2 or 0.10 for now if that makes life easier. Then it will be a single PR to change only this once we have all validated it in Jan/Feb.

Protocol.md
```diff
@@ -53,40 +55,9 @@ The following are features of Data and Metadata.
     - Connection via Metadata Push and Stream ID of 0
     - Individual Request or Response
 
-## Operation
-
-### Frame Header Format
+## Framing Protocol
```
benjchristensen: Good point, it is odd to have the parent and child headers with the same name.

Protocol.md
tmontgomery: I think it might be OK to leave this in and use 0. Removing it doesn't change much for the majority of frames.

Protocol.md
tmontgomery: I'll use 0.2 until we are happy with everything.

Protocol.md
```diff
@@ -99,17 +70,73 @@ have no further frames sent and all frames will be ignored.
...
 | Aeron                          | __NO__  |
+| HTTP/2                         | __NO__  |
 | Other                          | __YES__ |
 
+### Framing Header Format
+
+When using a transport protocol providing framing, the ReactiveSocket frame is simply encapsulated into the transport protocol messages directly.
+
+```
+    +-----------------------------+-----------------------
```
tmontgomery: Yes, it is. Let me see if it can be cleaner.

Protocol.md
tmontgomery: I think this is the only case that shows up.... but if I've missed any, let me know.

NiteshKant: I should have been clearer. What I was saying is that maybe we can trim this diagram to 24 bits. If we are consistently using 32 bits for all diagrams then `Framing Format` diagram is an outlier.

Protocol.md
```diff
@@ -221,11 +255,13 @@ Frame Contents
      * (__M__)etadata: Metadata present
      * (__L__)ease: Will honor LEASE (or not).
      * (__S__)trict: Adhere to strict interpretation of Data and Metadata.
+     * (__R__)esume Enable: Client requests resume capability is possible (Optional).
 * __Major Version__: (16) Major version number of the protocol.
 * __Minor Version__: (16) Minor version number of the protocol.
 * __Time Between KEEPALIVE Frames__: Time (in milliseconds) between KEEPALIVE frames that the client will send.
 * __Max Lifetime__: Time (in milliseconds) that a client will allow a server to not respond to a KEEPALIVE before
 it is assumed to be dead.
+* __Resume Identification Token__: (128) Token used for client resume identification. (Optional - set to all 0s if not supported)
```
benjchristensen: On the question about whether 128 bits is sufficient, do we want to land this PR and then discuss that separately? For example, should this support variable length?

Or are we comfortable at this point with the use of 128 bits, and the fact that we can always change it in future versions?

Protocol.md
tmontgomery: You are correct. I will change that. The original text mentions merging for this reason.

Protocol.md
tmontgomery: I would say let's land this and see if the need to make it variable comes up.

Protocol.md
```diff
@@ -99,17 +70,73 @@ have no further frames sent and all frames will be ignored.
 
 ReactiveSocket as specified here only allows for TCP, WebSocket, and Aeron as transport protocols.
```
tmontgomery: I wouldn't bother with H2 streams in this document. It's going to be a deeper subject and best handled in its own doc.

Protocol.md
benjchristensen: Works for me.

Protocol.md
benjchristensen: Then it is somewhat confusing to then reference HTTP/2 in the later section about framing. Should that later mention be removed?

benjchristensen:
I suggest we merge and then make further revisions in small PRs. Most things have been fixed I think.

Otherwise this will become very unruly.

Protocol.md
benjchristensen: If this is still an issue, we can fix in a subsequent PR after merging this.

benjchristensen:
Discuss with Todd. Merging so we can move forward. Can always revert if needed but would rather start working on smaller diffs.

PR #129, opened 2016-12-22
Supported Transports

benjchristensen:
The current version of Protocol.md states this:

> ReactiveSocket as specified here only allows for TCP, WebSocket, and Aeron as transport protocols.

Then in the next section, a couple paragraph away it mentions HTTP/2 as needing frame length.

It is confusing to say the protocol only supports TCP, WebSocket, and Aeron, and then mention "HTTP/2" and "Other" in the framing section. 

In this change I propose a possible wording to resolve the confusion. But I'm not quite sure what was meant by this sentence, when the section itself is also permitting other transports. 

So what is trying to be said, and what can we do to make this clearer?

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -68,7 +68,7 @@ The ReactiveSocket protocol uses a lower level transport protocol to carry React
 An implementation MAY "close" a transport connection due to protocol processing. When this occurs, it is assumed that that connection will
 have no further frames sent and all frames will be ignored.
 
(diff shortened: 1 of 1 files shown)


NiteshKant:
Java implementation does not have a WebSocket transport anymore as we (Netflix) is not using that transport anymore. So, I would say remove WebSocket also or remove this sentence completely. IMO, it does not add much value as we are defining the transport properties we expect.

tmontgomery:
This works for me.

PR #134, opened 2016-12-22
Permit Transport Protocols Managing StreamId

benjchristensen:
Adding text as suggested by @tmontgomery that permits an implementation to move management of streamId to the underlying transport.


NiteshKant:
LGTM, I will wait for @tmontgomery to approve.

PR #135, opened 2016-12-22
Resume behavior (rebased)

tmontgomery:
Rebased version of the resume behavior.


benjchristensen:
Nitesh, I'm good with having this merged. Then we can let final validation happen over the next 2 months as we test the implementation before calling it 1.0.

benjchristensen:
@NiteshKant @stevegury @robertroeser Any chance this can be reviewed soon? If possible I'd like to merge it this week.

Protocol.md
```diff
@@ -273,6 +273,8 @@ rules MAY be used for handling layout. For example, `application/x.netflix+cbor`
 * __Setup Data__: includes payload describing connection capabilities of the endpoint sending the
 Setup header.
 
+__NOTE__: A server that receives a SETUP frame that has (__R__)esume Enabled set, but does not support resuming operation, must reject the SETUP with an ERROR.
```
benjchristensen: How is returning RESUME_NOT_SUPPORTED different than returning CONNECTION_ERROR_NO_RETRY? In either case the ReactiveSocket either gives up or reconnects with a SETUP frame.

Protocol.md
```diff
@@ -312,6 +314,7 @@ The Error Data is typically an Exception message, but could include stringified
 | __UNSUPPORTED_SETUP__          | 0x00000002 | Some (or all) of the parameters specified by the client are unsupported by the server. Stream ID MUST be 0. |
 | __REJECTED_SETUP__             | 0x00000003 | The server rejected the setup, it can specify the reason in the payload. Stream ID MUST be 0. |
 | __CONNECTION_ERROR__           | 0x00000101 | The connection is being terminated. Stream ID MUST be 0. |
+| __CONNECTION_ERROR_NO_RETRY__  | 0x00000102 | The connection is being terminated. May __NOT__ be resumed. Stream ID MUST be 0. |
```
benjchristensen: That's a good idea. I like __CONNECTION_ERROR_RESUME_REJECTED__ and possible also __CONNECTION_ERROR_RESUME_UNSUPPORTED__

Protocol.md
```diff
@@ -874,3 +877,146 @@ assume it is set and act accordingly.
...
+
+Client lifetime management for servers must be extended to incorporate the length of time a client may successfully attempt resumption passed a transport disconnect. The means of client lifetime management are totally up to the implementation.
+
+### Resume Operation
+
+All ERROR frames sent MUST be CONNECTION_ERROR or CONNECTION_ERROR_NO_RETRY error code.
+
+Client side resumption operation starts when the client desires to try to resume and starts a new transport connection. The operation then proceeds as the following:
+
+* Client sends RESUME frame. The client must NOT send any other frame types until resumption succeeds. The RESUME Identification Token MUST be the token used in the original SETUP frame. The RESUME Last Received Position field MUST be the last successfully received implied position from the server.
```
benjchristensen: I'll let @tmontgomery answer this one.

Protocol.md
NiteshKant: `RESUME_NOT_SUPPORTED` will give the client an opportunity to continue without resumability which I would assume will happen if the client ever wants to connect to that server.
Since, resumability is optional, I would hope a client expecting resumability can work without resumability. In such a case, leaving it on the client to continue without resume is better than forcing reconnect.

tmontgomery: I am not totally happy with this either. In the normal flow, there is no positive ack of SETUP success. Which means that changing things for RESUME_ENABLED is a bit odd and unclean.

Another option is to let everything be silent and allow non-resumable servers to continue. But this means a nasty surprise for the client if/when it tries to resume.

The cleanest solution I can think of is to require the non-resuming servers to error back. Yes, it is against the requirement of preventing required operation. But it is the cleanest and lowest cost way to get good behavior, IMO.

Protocol.md
tmontgomery: I think we should only add specific errors if they will be easy for an implementation to differentiate and add logic for. Both of those errors are good, I think. Let me see about cleaning them up a little.

Protocol.md
tmontgomery: This is more of a simplification. If an implementation wants to do retransmission of frames, then sending the exact frames from the client sent must wait until an OK from the server so that the client knows what the server has seen and what it hasn't. New requests would broke this. So, forcing a client to wait is cleaner.

Protocol.md
```diff
@@ -874,3 +877,146 @@ assume it is set and act accordingly.
...
+* On receiving an ERROR frame, the client MUST NOT attempt resumption again if the error code was CONNECTION_ERROR_NO_RETRY.
+* On receiving a RESUME_OK, the client:
+    * MUST assume that the next REQUEST, CANCEL, ERROR, and RESPONSE frames have an implied position commencing from the last implied positions
+    * MAY retransmit *all* REQUEST, CANCEL, ERROR, and RESPONSE frames starting at the RESUME_OK Last Received Position field value from the server.
+    * MAY send an ERROR frame indicating the end of the connection and MUST NOT attempt resumption again
+
+Server side resumption operation starts when the client sends a RESUME frame. The operation then proceeds as the following:
+
+* On receiving a RESUME frame, the server:
+    * MUST send an ERROR frame if the server does not support resuming operation. This is accomplished by handling the Ignore bit in the RESUME frame.
```
tmontgomery: I've struggled with how to word this. Essentially, if a server doesn't support resuming, but receives a RESUME frame, then the I bit (ignore) in the frame would be set to 0. Which means the server should handle it like any frame it doesn't understand with the I bit set. I.e. terminate connection with a connection error.

Protocol.md
yschimke: I question the value of not having a SETUP ACK, mainly because we have specified typical keepalives are subsecond.  A client that wants to connect to a server and just send metadata push and fireAndForget could be talking to a brick wall for all it knows.

tmontgomery: We did extensively discuss this at one point. There is a little bit of a contention between trying to do a false start type of approach and having a SETUP ACK. The compromise was the LEASE behavior, which needed an ACK.

NiteshKant: Ok, so looks like there are these options:

- Explicit ack for setup (suggesting resume supported/unsupported and potentially some other optional setup properties)
- Reject setup if resume not supported by server.
- "Soft error" for resume not supported by server and let client decide whether to continue or disconnect. This is not

IMO, the third option is a good fit and does not preclude false start. Do you think it has other drawbacks?

Protocol.md
NiteshKant: My suggestion was to just rename `CONNECTION_ERROR_NO_RETRY` to `RESUME_FAILED` as that I believe is the intent. `NO_RETRY` is kind of misleading as to why would someone not retry a connect.

Protocol.md
NiteshKant: Ok. If I understand it correctly, if we were to allow new requests without waiting for `RESUME_OK` then the resumption state must be maintained per stream. Failure to do that may resend responses for the requests made between `RESUME` and `RESUME_OK`. Is my understanding correct?

Protocol.md
NiteshKant: Got it! I think it is confusing as the frame definition comes later in the spec 😄 
Maybe we can just say that `RESUME` frame can not be ignored as suggested by the Ignore flag (set to 0) in the frame.

benjchristensen:
@NiteshKant you mentioned on Gitter that you're ready to merge this and allow any small remaining things to be discussed separately. May I merge?

benjchristensen:
Thanks Nitesh!

PR #137, opened 2017-01-11
Add RESPONSE Next Flag

benjchristensen:
As per discussion in https://github.com/ReactiveSocket/reactivesocket/issues/126 in order to remove ambiguity and use of sentinel value of 0-bytes, and allow 0-byte payloads.

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -557,7 +557,7 @@ Frame Contents
     * (__N__)ext: bit to indicate Next (Response Data).
        * If set, `onNext` will be invoked on Subscriber/Observer.
     * (__C__)omplete: bit to indicate COMPLETE.
-       * If set, `onComplete` will be invokved on Subscriber/Observer.
+       * If set, `onComplete` will be invoked on Subscriber/Observer.
 * __Response Data__: payload for Reactive Streams onNext.
 
 A Response is generally referred to as a NEXT.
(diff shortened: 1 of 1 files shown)


Protocol.md
```diff
@@ -546,15 +546,18 @@ Frame Contents
...
                          Metadata & Response Data
 ```
 
 * __Flags__:
     * (__M__)etadata: Metadata Present.
     * (__F__)ollows: More fragments follow this fragment.
+    * (__N__)ext: bit to indicate Next (Response Data).
+       * If set, `onNext` will be invoked on Subscriber/Observer.
     * (__C__)omplete: bit to indicate COMPLETE.
+       * If set, `onComplete` will be invokved on Subscriber/Observer.
```
yschimke: typo

Protocol.md
```diff
@@ -546,15 +546,18 @@ Frame Contents
...
-    |   Frame Type  |0|M|F|C| Flags |
+    |   Frame Type  |0|M|F|N|C| ... |
     +-------------------------------+-------------------------------+
                          Metadata & Response Data
 ```
 
 * __Flags__:
     * (__M__)etadata: Metadata Present.
     * (__F__)ollows: More fragments follow this fragment.
+    * (__N__)ext: bit to indicate Next (Response Data).
```
yschimke: Should this always be set even when the frame contains data and/or metadata?

benjchristensen: Metadata is delivered via `onNext(Payload p)`, so my intent was yes.

benjchristensen: Added clarification to text. Is that better?

benjchristensen:
Thanks for the review Nitesh. 

@tmontgomery are you okay with this change and my wording?

tmontgomery:
I am good with this.

PR #138, opened 2017-01-12
correct stream id logic

yschimke:
From the HTTP/2 Spec

Streams are identified with an unsigned 31-bit integer.  Streams
   initiated by a client MUST use odd-numbered stream identifiers; those
   initiated by the server MUST use even-numbered stream identifiers.  A
   stream identifier of zero (0x0) is used for connection control
   messages; the stream identifier of zero cannot be used to establish a
   new stream.

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -182,7 +182,7 @@ Stream ID value of 0 is reserved for any operation involving the connection.
 A stream ID must be locally unique for a Requester in a connection.
 
 Stream ID generation follows general guidelines for [HTTP/2](https://tools.ietf.org/html/rfc7540) with respect
-to odd/even values. In other words, a client MUST generate even Stream IDs and a server MUST generate odd Stream IDs.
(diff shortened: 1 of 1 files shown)


yschimke:
The updated logic matches reactivesocket-java

https://github.com/ReactiveSocket/reactivesocket-java/blob/272401b1cbe2ea5f6bc7a16d4af18900c43a6f75/reactivesocket-core/src/main/java/io/reactivesocket/StreamIdSupplier.java

yschimke:
https://github.com/ReactiveSocket/reactivesocket/pull/92/files

yschimke:
But reactivesocket-cpp matches the current spec

https://github.com/ReactiveSocket/reactivesocket-cpp/blob/ff931c00c533d0db1d213d8503fc5435112e6085/src/StandardReactiveSocket.cpp#L60

yschimke:
cc @benjchristensen

PR #139, opened 2017-01-20
Make Resume Identification Token Variable Length

benjchristensen:
Support up to 2^16 bytes length for token

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -236,13 +236,7 @@ Frame Contents
     +---------------------------------------------------------------+
     |                         Max Lifetime                          |
     +---------------------------------------------------------------+
-    |                    Resume Identification Token                |
-    +                                                               +
-    |                                                               |
-    +                                                               +
-    |                                                               |
-    +                                                               +
-    |                                                               |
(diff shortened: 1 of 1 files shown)


Protocol.md
```diff
@@ -261,7 +255,8 @@ Frame Contents
 * __Time Between KEEPALIVE Frames__: Time (in milliseconds) between KEEPALIVE frames that the client will send.
 * __Max Lifetime__: Time (in milliseconds) that a client will allow a server to not respond to a KEEPALIVE before
 it is assumed to be dead.
-* __Resume Identification Token__: (128) Token used for client resume identification. (Optional - set to all 0s if not supported)
+* __Resume Identification Token Length__: Resume Identification Token Length in bytes. (Default to 0 if resumption not being used)
```
yschimke: The convention used for frame length is nice

_(24 = max 16,777,216 bytes)_

or for here

_(16 = max 65,536 bytes)_

benjchristensen: I'll change that.

benjchristensen:
Thanks for the review and approval.

PR #140, opened 2017-01-20
Frame Type from 8 to 6 bits to increase Flags

benjchristensen:
Move 2 bits from Frame Type to Flags.

Discussions have led to the conclusion that we don't need more that 2^6 (64) frame types, but want to grant more room for future additions of flags.


tmontgomery:
EXT should be 0x3F (all 1s).

benjchristensen:
Right you are.

benjchristensen:
@NiteshKant and @stevegury Are you okay with this change?

benjchristensen:
Thanks for the review and "heart" @robertroeser Will merge since we have approval from @tmontgomery @robertroeser and Vijayan and myself.

PR #148, opened 2017-01-25
`CONNECTION_CLOSE` error code

NiteshKant:
As discussed in issue #59 having an error code to indicate "safe closure" similar to GO_AWAY frame in HTTP/2 will be useful to indicate graceful shutdown as opposed to a fatal error.

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -309,7 +309,7 @@ The Error Data is typically an Exception message, but could include stringified 
 | __INVALID_SETUP__              | 0x00000001 | The Setup frame is invalid for the server (it could be that the client is too recent for the old server). Stream ID MUST be 0. |
 | __UNSUPPORTED_SETUP__          | 0x00000002 | Some (or all) of the parameters specified by the client are unsupported by the server. Stream ID MUST be 0. |
 | __REJECTED_SETUP__             | 0x00000003 | The server rejected the setup, it can specify the reason in the payload. Stream ID MUST be 0. |
-| __CONNECTION_ERROR__           | 0x00000101 | The connection is being terminated. Stream ID MUST be 0. Sender or Receiver of this frame MAY not wait for outstanding streams to terminate before closing the connection.|
+| __CONNECTION_ERROR__           | 0x00000101 | The connection is being terminated. Stream ID MUST be 0. Sender or Receiver of this frame MAY close the connection immediately without waiting for outstanding streams to terminate.|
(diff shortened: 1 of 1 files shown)


Protocol.md
```diff
@@ -309,8 +309,9 @@ The Error Data is typically an Exception message, but could include stringified
 | __INVALID_SETUP__              | 0x00000001 | The Setup frame is invalid for the server (it could be that the client is too recent for the old server). Stream ID MUST be 0. |
 | __UNSUPPORTED_SETUP__          | 0x00000002 | Some (or all) of the parameters specified by the client are unsupported by the server. Stream ID MUST be 0. |
 | __REJECTED_SETUP__             | 0x00000003 | The server rejected the setup, it can specify the reason in the payload. Stream ID MUST be 0. |
-| __CONNECTION_ERROR__           | 0x00000101 | The connection is being terminated. Stream ID MUST be 0. |
+| __CONNECTION_ERROR__           | 0x00000101 | The connection is being terminated. Stream ID MUST be 0. Sender or Receiver of this frame MAY not wait for outstanding streams to terminate before closing the connection.|
```
benjchristensen: The "MAY not wait" part required me re-reading a couple times as it comes across as a double negative for me. Perhaps clearer worded as follows?

> Sender or Receiver of this frame MAY close the connection immediately without waiting for outstanding streams to terminate.

Protocol.md
```diff
@@ -309,8 +309,9 @@ The Error Data is typically an Exception message, but could include stringified
...
 | __CONNECTION_ERROR_NO_RETRY__  | 0x00000102 | The connection is being terminated. May __NOT__ be resumed. Stream ID MUST be 0. |
+| __CONNECTION_CLOSE__           | 0x00000103 | The connection is being terminated. Stream ID MUST be 0. Sender or Receiver of this frame MUST wait for outstanding streams to terminate before closing the connection. New requests MAY not be accepted.|
```
benjchristensen: What is the reason for allowing new requests to potentially be accepted during shutdown?

Protocol.md
NiteshKant: Makes sense, will modify the language.

Protocol.md
NiteshKant: Not a strong reason but I thought this is a good suggestion rather than a rule that new requests MUST not be accepted, which may require much more work for an implementation. I am fine changing it to "New requests MUST not be accepted"

benjchristensen:
I'm good with this.

benjchristensen:
@tmontgomery Can you also confirm this? 
@vjn as well?

vjn:
LGTM

PR #151, opened 2017-01-26
Make Resume Token Not Present If Flag Not Set

benjchristensen:
This makes the behavior for resume token similar to metadata. If the flag is not set, then the bytes will not exist in the payload.

This is discussed in https://github.com/ReactiveSocket/reactivesocket/issues/149


vjn:
Nit: the max size for the token is 65,535 bytes.

tmontgomery:
I like this also. Good to go.

PR #152, opened 2017-01-26
REJECTED_RESUME vs CONNECTION_ERROR_NO_RETRY

benjchristensen:
Replace CONNECTION_ERROR_NO_RETRY with RESUME_REJECTED to be clearer.

The REJECTED_RESUME frame mirrors the REJECTED_SETUP error.

If SETUP fails, then REJECTED_SETUP. If RESUME fails, then REJECTED_RESUME.

This was discussed in https://github.com/ReactiveSocket/reactivesocket/issues/147


tmontgomery:
This was my original intention. Thanks for clarifying it.

benjchristensen:
If someone approves then, I'll merge.

PR #154, opened 2017-01-27
Request Channel Payloads from Requester

benjchristensen:
Clarifying that a channel sends PAYLOAD frames. 

- Depends on https://github.com/ReactiveSocket/reactivesocket/pull/153 and will need rebasing. 
- Need to also change wording later in the document explaining interaction model.
- Not certain about wording yet.


benjchristensen:
@tmontgomery I'd appreciate your help getting the wording correct.

vjn:
Some questions:

- Why is the N bit not also present for REQUEST_STREAM?
- The reference to 2^63-1 semantics from reactivestreams is made in both initial-request-n cases; but not for the REQUEST_N frame. Shouldn't it also be added there/moved to a global place?
- Alternatively, do we want to have those same semantics (don't track demand), but for 2^32-1?

benjchristensen:
> Alternatively, do we want to have those same semantics (don't track demand), but for 2^31-1?

I'd rather not. Honestly, over a network boundary I think the values should all be small-ish in almost all cases. And the performance benefit of the "don't track demand" thing just doesn't really matter when network is involved, whereas it has a slight benefit in-process. That's the only reason that exception was added.

benjchristensen:
> The reference to 2^63-1 semantics from reactivestreams is made in both initial-request-n cases; but not for the REQUEST_N frame. Shouldn't it also be added there/moved to a global place?

Where do you suggest putting it?

benjchristensen:
> Why is the N bit not also present for REQUEST_STREAM?

Why do we even have that N-bit? @tmontgomery what was the thinking on that?

NiteshKant:
>>> Why do we even have that N-bit?

I think it is because request-channel frames can be sent multiple times (before we changed it in #154) for the same request but they may not necessarily have request-n. Since, we have changed that interaction now to only send request-channel once, that N-bit does not make sense and we can always have initial request-n

benjchristensen:
K, good. I'll remove that.

benjchristensen:
Should be fixed now.

vjn:
Since we have the reference to "2^63-1 semantics from ReactiveStreams" in three places now, perhaps that could be combined and moved to the "Flow Control" section?

benjchristensen:
Trying again ... how about this?

tmontgomery:
An N-bit in REQUEST_STREAM doesn't make sense, I don't think.

benjchristensen:
@tmontgomery You good with these as well?

tmontgomery:
Good for me.

PR #156, opened 2017-02-10
Added client's "first-available" position in RESUME frame.

vjn:
Sending both an expectation on the server's retention of the client's last-received, and an offer of what positions the client has retained lets the server immediately determine whether resumption can/should be allowed.

The alternative would be for the server to restart half of a resumption, but for the resumption to be later failed, when the RESUME_OK is received by the client.


Protocol.md
```diff
@@ -983,7 +989,8 @@ RESUME frames MUST always use Stream ID 0 as they pertain to the connection.
 * __Resume Identification Token__: (128) Token used for client resume identification. Same Resume Identification used in the initial SETUP by the client.
 * __Resume Identification Token Length__: (16 = max 65,536 bytes) Resume Identification Token Length in bytes. 
 * __Resume Identification Token__: TToken used for client resume identification. Same Resume Identification used in the initial SETUP by the client.
-* __Last Received Position__: (64) The last implied position the client received from the server
+* __Last Received Server Position__: (64) The last implied position the client received from the server.
+* __First Available Client Position__: (65) The earliest position that the client can rewind back to prior to resending frames.
```
benjchristensen: Why does this show (65) bytes? Typo?

benjchristensen:
This change makes sense to me.

Protocol.md
benjchristensen: And by bytes I meant bits.

benjchristensen:
@tmontgomery I'd appreciate your review of this.

tmontgomery:
Aside from my comment above about the resend language, I am fine.

benjchristensen:
@tmontgomery I don't see another comment from you.

tmontgomery:
The comment above should be showing up and is still relevant.

"I think we should retain the ability to do resumption without retransmit. So, could the wording be changed to say "both sides (client and server) may need to resend frames..." and "Now both the client and the server each may re-send frames as necessary.".

Another option is to reword without using "resend", etc."

By tying in the "resend" language, we are requiring resumption to retransmit missing frames. This is not ideal, IMO.

benjchristensen:
> The comment above should be showing up and is still relevant.

I don't see a comment from you. Can you submit a PR with the changes?

tmontgomery:
Will do.

PR #157, opened 2017-02-14
Version in RESUME Frame

benjchristensen:
Adding version in SETUP frame as we realized while implementing v1.0 spec changes that we don't know what version of frames are coming in when we receive a SETUP frame.


lehecka:
this is what I asked for, so it LGTM!

PR #158, opened 2017-02-14
Clarify Payload vs Response

benjchristensen:
We changed "Response" to "Payload" previously, but not all wording had been updated. Attempting to cleanup and clarify here.


benjchristensen:
@fiorix This is part of some cleanup that should help clarity. Does this improve things? Anything else to make this clearer?

fiorix:
👍

PR #160, opened 2017-02-14
Match Frame Sections to Frame Names

benjchristensen:
Based on feedback of people reading spec for first time, changing section names to match the frame names, and putting in the specific "Frame Type" information for each frame with their hex value.

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -613,7 +613,7 @@ The general format for an extension frame is given below.
                           Depends on Extended Type...
 ```
 
-* __Frame Type__: (16) 0xFFFF
+* __Frame Type__: (16) 0x3F
 * __Flags__:
     * (__I__)gnore: Can be frame be ignored if not understood?
     * (__M__)etadata: Metadata Present.
--- a/Protocol.md
+++ b/Protocol.md
@@ -204,7 +204,7 @@ to odd/even values. In other words, a client MUST generate odd Stream IDs and a 
 | __RESUME_OK__                  | 0x0E | __Resume OK__ : Sent in response to a RESUME if resuming operation possible (optional) |
 | __EXT__                        | 0x3F | __Extension Header__: Used To Extend more frame types as well as extensions. |
 
-### Setup Frame
+### SETUP Frame (0x01)
 
 Setup frames MUST always use Stream ID 0 as they pertain to the connection.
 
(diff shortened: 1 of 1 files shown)


tmontgomery:
RESUME_OK might be missing its hex value.

benjchristensen:
It was there already, but formatted differently, so I updated the formatting to match.

benjchristensen:
@tmontgomery Does the fix I made look okay? If so, can you approve this change?

benjchristensen:
@fiorix Some of these changes are based on feedback from you. Does this improve the clarity of the doc? Any other changes you suggest?

Protocol.md
```diff
@@ -613,7 +613,7 @@ The general format for an extension frame is given below.
                           Depends on Extended Type...
 ```
 
-* __Frame Type__: (16) 0xFFFF for Extension Header.
+* __Frame Type__: (16) 0xFFFF
```
fiorix: Is 0xFFFF correct or should it be 0x3F?

tmontgomery: I think this is 0x3F

Protocol.md
```diff
@@ -995,7 +1007,7 @@ RESUME frames MUST always use Stream ID 0 as they pertain to the connection.
 * __Last Received Server Position__: (64) The last implied position the client received from the server.
 * __First Available Client Position__: (64) The earliest position that the client can rewind back to prior to resending frames.
 
-#### Resume OK Frame
+#### RESUME_OK Frame
```
tmontgomery: This should add the hex value 0x0E

tmontgomery:
Some nits picked.

Protocol.md
benjchristensen: Hmm, that seems to have regressed in merges along the way. I'm positive we fixed that a few weeks back. Fixing again ... thanks for flagging it.

benjchristensen: Actually, that was a different place. The other place was fixed. Not sure how this was still wrong :-) Anyhow ... fixed now.

Protocol.md
benjchristensen: Done.

benjchristensen:
Changes made.

tmontgomery:
awesome!

PR #166, opened 2017-02-17
Fix "bytes" to be "bits"

moroshko:


Protocol.md
```diff
@@ -167,7 +167,7 @@ If Metadata Length is greater than this value, the entire frame MUST be ignored.
     +---------------------------------------------------------------+
 ```
 
-* __Metadata Length__: (24 = max 16,777,216 bytes) unsigned positive Integer representing the length of Metadata in bytes. Excluding Metadata Length field.
```
benjchristensen: I think the confusion is the first use of bytes. We should just remove that word. But the second use is supposed to be bytes. 

So:

(24 = max 16,777,216) Unsigned ...

That max number is what can exist in 24-bits. That number then represents how many bytes long the payload/frame/whatever can be.

benjchristensen:
When we discussed earlier I miscommunicated with you. These numbers represent the number of bytes in length.

moroshko:
My gut feel was telling me that bytes make more sense 😄

PR #167, opened 2017-02-17
Remove M flag from KEEPALIVE Frame

moroshko:
Since there is no `M` in the Frame Contents visualization, I suggest we don't mention it at all.

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -517,8 +517,6 @@ Frame Contents
 ```
 
 * __Frame Type__: (16) 0x08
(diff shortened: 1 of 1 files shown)


benjchristensen:
@tmontgomery What do you recommend? If we decide to keep it, then we should probably be consistent and identify all flags for all frames that are pre-defined.

tmontgomery:
Removing it is fine. Just make sure to have the bit set to 0 in the header location.

PR #171, opened 2017-02-17
Swap Complete and Next flags for PAYLOAD

lexs:
This means the position of the Complete flag matches other frames which reduces confusion when implementing the spec.

cc @benjchristensen


benjchristensen:
Yes, this makes it match REQUEST_CHANNEL.

PR #172, opened 2017-02-17
Put (M)etdata present flag before Resume enabled

lexs:
This makes it consistent with what the frame header section mentions.

cc @benjchristensen


benjchristensen:
Good catch.

PR #174, opened 2017-02-17
Hardcode Stream ID to 0 in diagrams when known

moroshko:
Similarly to specifying the known flags values in the diagrams, I suggest we do the same for Stream ID. 

I find it easier to consume the diagram when I see hardcoded values, because it suggests I need to focus on the other fields.


benjchristensen:
@tmontgomery what do you think? Seems reasonable to me.

tmontgomery:
Agreed.

PR #175, opened 2017-02-17
Detailed and Consistent Bits/Max Size

benjchristensen:
Attempting to be very clear and precise, to the point of pedantic, so there are no questions or misinterpretations as people read the spec.

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -62,7 +62,7 @@ The following are features of Data and Metadata.
 The ReactiveSocket protocol uses a lower level transport protocol to carry ReactiveSocket frames. A transport protocol MUST provide the following:
 
 1. Unicast [Reliable Delivery](https://en.wikipedia.org/wiki/Reliability_(computer_networking)).
-1. [Connection-Oriented](https://en.wikipedia.org/wiki/Connection-oriented_communication) and preservation of frame ordering. Frame A sent before Frame B must arrive in source order. i.e. if Frame A is sent by the same source as Frame B, then Frame A will always arrive before Frame B. No assumptions about ordering across sources is assumed.
+1. [Connection-Oriented](https://en.wikipedia.org/wiki/Connection-oriented_communication) and preservation of frame ordering. Frame A sent before Frame B MUST arrive in source order. i.e. if Frame A is sent by the same source as Frame B, then Frame A will always arrive before Frame B. No assumptions about ordering across sources is assumed.
 1. [FCS](https://en.wikipedia.org/wiki/Frame_check_sequence) is assumed to be in use either at the transport protocol or at each MAC layer hop. But no protection against malicious corruption is assumed.
 
 An implementation MAY "close" a transport connection due to protocol processing. When this occurs, it is assumed that the connection will have no further frames sent and all frames will be ignored.
@@ -71,7 +71,7 @@ ReactiveSocket as specified here has been designed for and tested with TCP, WebS
 
 ### Framing Protocol Usage
 
-Some of the supported transport protocols for ReactiveSocket may not support specific framing that preserves message boundaries. For these protocols, a framing protocol must be used with the ReactiveSocket frame that prepends the ReactiveSocket Frame Length.
+Some of the supported transport protocols for ReactiveSocket may not support specific framing that preserves message boundaries. For these protocols, a framing protocol MUST be used with the ReactiveSocket frame that prepends the ReactiveSocket Frame Length.
 
 The frame length field MUST be omitted if the transport protocol preserves message boundaries e.g. provides compatible framing. If, however, the transport protocol only provides a stream abstraction or can merge messages without preserving boundaries, or multiple transport protocols may be used, then the frame header MUST be used.
 
@@ -93,7 +93,7 @@ When using a transport protocol providing framing, the ReactiveSocket frame is s
     +-----------------------------------------------+
 ```
 
-When using a transport protocol that does not provide compatible framing, the Frame Length must be prepended to the ReactiveSocket Frame.
+When using a transport protocol that does not provide compatible framing, the Frame Length MUST be prepended to the ReactiveSocket Frame.
 
 ```
      0                   1                   2
@@ -130,7 +130,7 @@ ReactiveSocket frames begin with a ReactiveSocket Frame Header. The general layo
   * Transport protocols that include demultiplexing, such as HTTP/2, MAY omit the Stream ID field if all parties agree. The means of negotiation and agreement is left to the transport protocol. 
 * __Frame Type__: (6 bits = max value 63) Type of Frame.
 * __Flags__: (10 bits) Any Flag bit not specifically indicated in the frame type should be set to 0 when sent and not interpreted on
-reception. Flags generally depend on Frame Type, but all frame types must provide space for the following flags:
+reception. Flags generally depend on Frame Type, but all frame types MUST provide space for the following flags:
      * (__I__)gnore: Ignore frame if not understood
      * (__M__)etadata: Metadata present
 
@@ -177,7 +177,7 @@ the semantics of the stream based on its type.
 
 Stream ID value of 0 is reserved for any operation involving the connection.
 
-A stream ID must be locally unique for a Requester in a connection.
+A stream ID MUST be locally unique for a Requester in a connection.
 
 Stream ID generation follows general guidelines for [HTTP/2](https://tools.ietf.org/html/rfc7540) with respect
 to odd/even values. In other words, a client MUST generate odd Stream IDs and a server MUST generate even Stream IDs.
@@ -268,7 +268,7 @@ rules MAY be used for handling layout. For example, `application/x.netflix+cbor`
 * __Setup Data__: includes payload describing connection capabilities of the endpoint sending the
 Setup header.
 
-__NOTE__: A server that receives a SETUP frame that has (__R__)esume Enabled set, but does not support resuming operation, must reject the SETUP with an ERROR. 
+__NOTE__: A server that receives a SETUP frame that has (__R__)esume Enabled set, but does not support resuming operation, MUST reject the SETUP with an ERROR. 
 
 ### ERROR Frame (0x0B)
 
@@ -930,7 +930,7 @@ When a client sends a RESUME frame, it sends two implied positions: the last fra
 
 ### Client Lifetime Management
 
-Client lifetime management for servers must be extended to incorporate the length of time a client may successfully attempt resumption passed a transport disconnect. The means of client lifetime management are totally up to the implementation.
+Client lifetime management for servers MUST be extended to incorporate the length of time a client may successfully attempt resumption passed a transport disconnect. The means of client lifetime management are totally up to the implementation.
 
 ### Resume Operation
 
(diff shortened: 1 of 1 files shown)


benjchristensen:
This is detailed enough that I'm going blind to things I'm missing. Please help get this right @tmontgomery @lehecka @lexs @moroshko @yschimke

Protocol.md
```diff
@@ -291,11 +291,12 @@ Frame Contents
                         Metadata & Error Data
 ```
 
-* __Frame Type__: (16) 0x0B
-* __Flags__:
+* __Frame Type__: (6 bits = max value 63) 0x0B
+* __Flags__: (8 bits)
```
lexs: Isn't this 10? 16 - 6?

benjchristensen: Yup :-/ *facepalm* here ... remember the going blind comment? fixing now.

Protocol.md
```diff
@@ -106,7 +106,7 @@ When using a transport protocol that does not provide compatible framing, the Fr
     +-----------------------------------------------+
 ```
 
-* __Frame Length__: (24 = max 16,777,216 bytes) Length of Frame. Excluding Framing Length Field.
+* __Frame Length__: (24 bits = max value 16,777,215) Positive integer representing the length of Frame in bytes. Excluding Framing Length Field.
```
moroshko: unsigned integer? 
(grep the spec for "integer", and make it explicit whether it's signed or unsigned)

Protocol.md
```diff
@@ -126,10 +126,10 @@ ReactiveSocket frames begin with a ReactiveSocket Frame Header. The general layo
     +-------------------------------+
 ```
 
-* __Stream ID__: (32) Positive signed integer representing the stream Identifier for this frame or 0 to indicate the entire connection.
+* __Stream ID__: (32 bits = max value 2^31-1 = 2,147,483,647) Positive signed integer representing the stream Identifier for this frame or 0 to indicate the entire connection.
```
moroshko: Remove "2^31-1 =" for consistency sake

Protocol.md
```diff
@@ -126,10 +126,10 @@ ReactiveSocket frames begin with a ReactiveSocket Frame Header. The general layo
     +-------------------------------+
 ```
 
-* __Stream ID__: (32) Positive signed integer representing the stream Identifier for this frame or 0 to indicate the entire connection.
+* __Stream ID__: (32 bits = max value 2^31-1 = 2,147,483,647) Signed positive Integer representing the stream Identifier for this frame or 0 to indicate the entire connection.
```
moroshko: Remove `2^31-1 =` for consistency?

benjchristensen:
Merged and resolved conflicts after other PRs landed, fixed the issue @lexs brought up, and did another pass with little tweaks.

Protocol.md
```diff
@@ -244,16 +244,16 @@ Frame Contents
...
      * (__R__)esume Enable: Client requests resume capability if possible. Resume Identification Token present.
      * (__L__)ease: Will honor LEASE (or not).
      * (__S__)trict: Adhere to strict interpretation of Data and Metadata.
-* __Major Version__: (16) Major version number of the protocol.
-* __Minor Version__: (16) Minor version number of the protocol.
-* __Time Between KEEPALIVE Frames__: Time (in milliseconds) between KEEPALIVE frames that the client will send.
-* __Max Lifetime__: Time (in milliseconds) that a client will allow a server to not respond to a KEEPALIVE before
+* __Major Version__: (16 bits = max value 65,535) Positive Integer of Major version number of the protocol.
+* __Minor Version__: (16 bits = max value 65,535) Positive Integer of Minor version number of the protocol.
+* __Time Between KEEPALIVE Frames__: (32 bits = max value 2^31-1 = 2,147,483,647) Signed positive Integer of Time (in milliseconds) between KEEPALIVE frames that the client will send.
```
moroshko: Remove `2^31-1 =` for consistency?

Protocol.md
```diff
@@ -244,16 +244,16 @@ Frame Contents
...
+* __Max Lifetime__: (32 bits = max value 2^31-1 = 2,147,483,647) Signed positive Integer of Time (in milliseconds) that a client will allow a server to not respond to a KEEPALIVE before
```
moroshko: Remove `2^31-1 =`?

Protocol.md
```diff
@@ -395,10 +396,10 @@ Frame Contents
                                   Data
 ```
 
-* __Frame Type__: (16) 0x03
-* __Flags__:
+* __Frame Type__: (6 bits = max value 63) 0x03
+* __Flags__: (10 bits)
-* __Last Received Position__: (64) Resume Last Received Position (optional. Set to all 0s when not supported.)
+* __Last Received Position__: (64 bits = max value 2^63-1) Signed positive Long of Resume Last Received Position (optional. Set to all 0s when not supported.)
```
moroshko: Is there a value to specify the max value? It feels not very useful in this case. How about we omit the max value everywhere?

Protocol.md
```diff
@@ -460,11 +461,11 @@ Frame Contents
...
 ```
 
-* __Frame Type__: (16) 0x06
-* __Flags__:
+* __Frame Type__: (6 bits = max value 63) 0x06
+* __Flags__: (10 bits)
     * (__F__)ollows: More fragments follow this fragment.
-* __Initial Request N__: 32-bit signed integer representing the initial request N value for the stream. Only positive values are allowed.
+* __Initial Request N__: (32 bits = max value 2^31-1 = 2,147,483,647) Signed positive Integer representing the initial request N value for the stream. Value must be > 0.
```
moroshko: Often, "must" appears as "MUST". Would be nice to make it consistent.

Protocol.md
```diff
@@ -486,12 +487,12 @@ Frame Contents
...
 
-* __Frame Type__: (16) 0x07
-* __Flags__:
+* __Frame Type__: (6 bits = max value 63) 0x07
+* __Flags__: (10 bits)
     * (__M__)etadata: Metadata present
     * (__F__)ollows: More fragments follow this fragment.
     * (__C__)omplete: bit to indicate COMPLETE.
-* __Initial Request N__: 32-bit signed integer representing the initial request N value for channel. Only positive values are allowed.
+* __Initial Request N__: (32 bits = max value 2^31-1 = 2,147,483,647) Signed positive Integer representing the initial request N value for channel. Value must be > 0.
```
moroshko: must => MUST

Protocol.md
```diff
@@ -516,8 +517,8 @@ Frame Contents
     +---------------------------------------------------------------+
 ```
 
-* __Frame Type__: (16) 0x08
-* __Request N__: 32-bit signed integer value of items to request. Only positive values are allowed.
+* __Frame Type__: (6 bits = max value 63) 0x08
+* __Request N__: (32 bits = max value 2^31-1 = 2,147,483,647) Signed positive Integer of value of items to request. Value must be > 0.
```
moroshko: Just grep for "must" ;)

Protocol.md
```diff
@@ -589,8 +590,8 @@ Frame Contents
                                 Metadata
 ```
 
-* __Frame Type__: (16) 0x0C
-* __Stream ID__: Must be 0 to pertain to the entire connection.
+* __Stream ID__: (32 bits = max value 2^31-1 = 2,147,483,647) Must be 0 to pertain to the entire connection.
```
moroshko: I suggest we don't mention hardcoded fields, so I would delete this line. See #167 and #174.

Protocol.md
benjchristensen: I put that in specifically for 32 and 64 bit because they hit the max sizes and want to be explicit.

Protocol.md
benjchristensen: I removed signed/unsigned on anything smaller than 32 as it doesn't matter. They fit within a 32-bit signed value in a programming language. Does saying unsigned here improve clarity that the full 24 bits are usable?

Protocol.md
benjchristensen: Comment as above. Being pedanticly specific with 32 and 64 bit values.

Protocol.md
benjchristensen: It's very very big. Hence 2^63-1

Protocol.md
benjchristensen: MUST is the standard for rules that are required. So yes we should be consistent for that.

Protocol.md
moroshko: I disagree that it doesn't matter.

A signed or unsigned notion has nothing to do with programming languages. It just tells us how to interpret the bits.

So, here for example, if we say it is signed, and we get `Frame Length = 2^24 - 1`, we know something is wrong because we got a negative Frame Length. If we say it is unsigned, there are no wrong/error values.

My main point here is, that we need to separate between:
* how to interpret the bits - we do this by adding "signed" or "unsigned"
* any constraints on the value - e.g. "it MUST be positive"

In this specific instance, I suggest:

```diff
- Positive integer representing the length of Frame in bytes.
+ Unsigned integer representing the length of Frame in bytes.
```

In other instances, I suggest:

Stream ID: 
```diff
- Positive signed integer
+ Signed integer. MUST be non-negative.
```

Metadata Length:
```diff
- unsigned positive Integer
+ unsigned integer
```

Initial Request N:
> 32-bit signed integer representing the initial request N value for the stream. Only positive values are allowed.

This is actually a great example! The only thing I'd consider changing here is to use the [RFC 2119](https://tools.ietf.org/html/rfc2119) language (e.g. "MUST" instead of "allowed").

Protocol.md
tmontgomery: This should be explicit as @benjchristensen states.

Protocol.md
tmontgomery: Indeed. MUST language would be good for this instance.

tmontgomery:
This looks good to me. But could use another few set of eyes to catch anything.

Protocol.md
benjchristensen: > My main point here is, that we need to separate between:

You've convinced me ... I'll put in the signed/unsigned terminology in those places as well.

benjchristensen: I've committed an update that tries to adjust to this style. See if I missed anything or if you still have recommendations for improvement.

Protocol.md
benjchristensen: I've changed all uses of 'must' to 'MUST'

benjchristensen:
@moroshko Please do another pass of this.

moroshko:
@benjchristensen Looks great!!

Just few little things:
1. **Frame Length:** (24 bits = max value 16,777,215) Unsigned integer representing the length of Frame in bytes. Excluding Framing Length Field.
    * I'd say: "Excluding the Frame Length field"

2. REQUEST_N Frame - **Request N:** (32 bits = max value 2^31-1 = 2,147,483,647) Signed integer of value of items to request. Value MUST be > 0.
    * "Signed integer representing the amount of items to request." would sound better to me

3. A positive non-0 Stream ID means the error pertains to a given stream.
    * I'd delete "non-0"

4. METADATA_PUSH Frame - **Stream ID:** (32 bits = max value 2^31-1 = 2,147,483,647) Value MUST == 0 to pertain to the entire connection.
    * For other Frames where Stream ID is 0, we say e.g. "Lease frames MUST always use Stream ID 0 as they pertain to the Connection.", and don't mention **Stream ID** when describing the fields. I recommend to do the same for the METADATA_PUSH Frame.

5. We usually capitalize "Stream ID", but there are few instances of "stream ID". I suggest we keep it always "Stream ID".

6. **Resume Identification Token Length:** (16 bits = max value 65,535) Unsigned integer of Resume Identification Token Length in bytes.
    * I suggest we match the highlighted field name to what we have in the diagram, i.e.: **Token Length:** (16 bits = max value 65,535) Unsigned integer of Resume Identification Token Length in bytes.
    * Note this appears twice in the docs.

7. We state that "Stream ID MUST be >= 0", and also say in quite a few places that "Stream ID MUST be non-0". I wonder if we should say "Stream ID MUST be > 0" instead?

benjchristensen:
> I suggest we match the highlighted field name to what we have in the diagram, i.e.: Token Length:

Not enough space in the diagram to include the full name.

moroshko:
> Not enough space in the diagram to include the full name.

When I look at the diagram, and read the various fields, I might ask myself "What does this field mean?".

Naturally, I'll be looking for an answer below the diagram.

There, we highlight the field names, **which makes it easier to find what I'm looking for.**

But, if the highlighted text doesn't match the field name in the diagram, it doesn't really help that it's highlighted.

Since we don't have enough space in the diagram to put the full name, I'd change it to be:

> **Token Length:** (16 bits = max value 65,535) Unsigned integer of Resume Identification Token Length in bytes.

Take it or leave it ;)

PR #184, opened 2017-02-22
Change application reserved error code range

lexs:
We now have codes up to 204, so application needs to start at 205.

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -349,7 +349,7 @@ The Error Data is typically an Exception message, but could include stringified 
 | __RESERVED__                   | 0xFFFFFFFF | __Reserved for Extension Use__ |
 
 __NOTE__: Values in the range of 0x0001 to 0x00FF are reserved for use as SETUP error codes. Values in the range of
-0x00101 to 0x001FF are reserved for connection error codes. Values in the range of 0x00205 to 0xFFFFFFFE are reserved for application layer
+0x00101 to 0x001FF are reserved for connection error codes. Values in the range of 0x00301 to 0xFFFFFFFE are reserved for application layer
 errors.
 
 <a name="frame-lease"></a>
--- a/Protocol.md
+++ b/Protocol.md
@@ -348,9 +348,7 @@ The Error Data is typically an Exception message, but could include stringified 
 | __INVALID__                    | 0x00000204 | The request is invalid. Stream ID MUST be > 0. |
 | __RESERVED__                   | 0xFFFFFFFF | __Reserved for Extension Use__ |
 
-__NOTE__: Values in the range of 0x0001 to 0x00FF are reserved for use as SETUP error codes. Values in the range of
-0x00101 to 0x001FF are reserved for connection error codes. Values in the range of 0x00301 to 0xFFFFFFFE are reserved for application layer
-errors.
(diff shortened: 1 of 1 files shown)


lexs:
Should we bump this number to 300 to allow us to add more error codes?

Protocol.md
```diff
@@ -349,7 +349,7 @@ The Error Data is typically an Exception message, but could include stringified
 | __RESERVED__                   | 0xFFFFFFFF | __Reserved for Extension Use__ |
 
 __NOTE__: Values in the range of 0x0001 to 0x00FF are reserved for use as SETUP error codes. Values in the range of
-0x00101 to 0x001FF are reserved for connection error codes. Values in the range of 0x00201 to 0xFFFFFFFE are reserved for application layer
+0x00101 to 0x001FF are reserved for connection error codes. Values in the range of 0x00205 to 0xFFFFFFFE are reserved for application layer
```
yschimke: I think the intent was that errors where in distinct blocks, somewhat like 2XX, 3XX, 4XX, 5XX in HTTP.

So your suggestion to go to 0x300 seems like the only sensible approach.

benjchristensen: 0x301 seems good to me.

benjchristensen:
@tmontgomery what do you recommend?

Protocol.md
tmontgomery: I agree. 0x301.

tmontgomery:
Saving some space is good. 0x301 should be OK.

lexs:
Updated to 0x301.

Protocol.md
```diff
@@ -349,7 +349,7 @@ The Error Data is typically an Exception message, but could include stringified
 | __RESERVED__                   | 0xFFFFFFFF | __Reserved for Extension Use__ |
 
 __NOTE__: Values in the range of 0x0001 to 0x00FF are reserved for use as SETUP error codes. Values in the range of
-0x00101 to 0x001FF are reserved for connection error codes. Values in the range of 0x00201 to 0xFFFFFFFE are reserved for application layer
+0x00101 to 0x001FF are reserved for connection error codes. Values in the range of 0x00301 to 0xFFFFFFFE are reserved for application layer
```
yschimke: This text currently leaves a gap e.g. doesn't describe 0x00201 to 0x002FF

tmontgomery: Good catch. 0x201 to 0x301 should be marked for later use in future versions.

yschimke: You've confused me more!  But it's late here.

tmontgomery: Sorry. Meant that those (if I recall correctly) are for space for use by later versions of the protocol. "Reserved for future use".

benjchristensen:
Is this ready for merging?

yschimke:
No, I think it should be changed to explain the 4 ranges  
0x0001 to 0x00FF - setup errors
0x0101 to 0x01FF - connection errors
0x0201 to 0x02FF - predefined application errors
0x0301 to 0xFFFFFFFE? - custom application errors

starting custom application errors at 0x0205 seems like it needlessly precludes the spec from ever defining more.

lexs:
Updated the NOTE wording, should be ready now.

PR #187, opened 2017-02-24
clarify HTTP/2 stream as transport with link

benjchristensen:
Since HTTP/2 is referenced later, and we target it, I'm calling it out in the transport section, but being specific to it being HTTP/2 streams that we target, as most HTTP/2 clients don't expose stream APIs.


benjchristensen:
@tmontgomery you okay with this wording change, or can you recommend a better way?

This comes from feedback about how we mention HTTP/2 in the later table, but not in the earlier sentence about what we've targeted and tested.

PR #189, opened 2017-02-24
Stream Lifetime

benjchristensen:
Clarifying lifetime of streams, and what happens when the max 2^31-1 is hit.


benjchristensen:
Hmm, @tmontgomery I just noticed this and would appreciate your guidance. 

> Once a stream has "terminated", the Stream ID can be "forgotten" by the Requester and Responder. An implementation MAY re-use an ID at this time, but it is recommended that an implementation not aggressively re-use IDs.

That is different than what we've discussed before. I prefer we not allow reuse.

benjchristensen:
We also have this later in the doc:

> 7. Stream ID values MAY be re-used after completion or error of a stream.

benjchristensen:
So ... do we allow reuse or not?

tmontgomery:
For clarity, let's not allow re-use. So, that section should be nixed that discusses it.

Protocol.md
```diff
@@ -741,22 +746,18 @@ The possible sequences with LEASE are below.
 
 ## Fragmentation And Reassembly
 
-PAYLOAD frames and all REQUEST frames may represent a large object and MAY need to be fragmented to fit within the Frame Data size. When this
-occurs, the __F__ flag indicates if more fragments follow the current frame (or not).
+PAYLOAD frames and all REQUEST frames may represent a large object and MAY need to be fragmented to fit within the Frame Data size. When this occurs, the __F__ flag indicates if more fragments follow the current frame (or not).
```
benjchristensen: Just removed a stray newline while I was here.

Protocol.md
```diff
@@ -741,22 +746,18 @@ The possible sequences with LEASE are below.
...
 
 ## Stream Sequences and Lifetimes
 
-Streams exists for a specific period of time. So an implementation may assume that Stream IDs are valid for a finite period of time. This period
-of time is bound by, at most, the lifetime of the underlying transport protocol connection lifetime. Beyond that, each interaction pattern imposes
-lifetime based on a sequence of interactions between Requester and Responder.
+Streams exists for a specific period of time. So an implementation may assume that Stream IDs are valid for a finite period of time. This period of time is bound by either a) the lifetime of the underlying transport protocol connection, or b) the lifetime of a session if resumability is used to extend the session across multiple transport protocol connections. Beyond that, each interaction pattern imposes lifetime based on a sequence of interactions between Requester and Responder.
```
benjchristensen: Added clarification on lifetime with and without resumability.

benjchristensen:
Do these wording changes work?

tmontgomery:
Works for me.

PR #194, opened 2017-02-24
Metadata before Data when fragmented

benjchristensen:
Require implementations to send Metadata before Data when fragmented. 

Proposed for https://github.com/ReactiveSocket/reactivesocket/issues/170


benjchristensen:
@tmontgomery you okay with this wording and example?

tmontgomery:
I am good.

PR #197, opened 2017-02-24
Use unsigned for all bit descriptions

benjchristensen:
Based on feedback, adopting the style of description used by HTTP/2 (http://httpwg.org/specs/rfc7540.html#FrameHeader) and specifically showing the first bit of 32-bit and 64-bit numbers to always be 0.

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -133,7 +133,7 @@ When using a transport protocol that does not provide compatible framing, the Fr
     +-----------------------------------------------+
 ```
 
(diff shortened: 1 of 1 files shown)


benjchristensen:
Does this do a better job of describing the bits and appropriate number sizes? Or should it be left as is?

tmontgomery:
This is a good change, IMO. Clearer for others, I think.

nickgg:
It looks clearer to me.

PR #200, opened 2017-02-28
KEEPALIVE timing suggestions

benjchristensen:
- moved into SETUP where the value is
- include server-to-server and mobile-to-server commentary


tmontgomery:
These seem OK values. But what has driven them as suggestions?

benjchristensen:
The 500ms value has been in the spec all along, I thought it came from you! :-) If it didn't, do you want to remove it? I was only leaving it in there as I thought it was from you and I'm okay with the value. I then added mobile so that it wasn't solely focused on server-to-server since it has confused people to see the 500ms value recommended when we're using it for mobile.

The mobile one comes from experience we have with battery. We actually use a keepalive higher than 30 seconds.

tmontgomery:
OK. Looking at the phrasing, I think it is fine. Yes, the original server-to-server was a suggestion. Was curious if any anecdotal experience had shown differently. :)

My intervals for timeouts have been dropping recently just due to timeout values dropping as well. So, my perception is somewhat skewed at the moment.

benjchristensen:
For the server, should we use SHOULD or something else? That is a strong suggestion. For mobile I left it very loose on purpose.

benjchristensen:
language softened

PR #203, opened 2017-02-28
clarify request(n) being cumulative

benjchristensen:
Add clarity about request(n) being cumulative.


tmontgomery:
This is a good clarification. But for more detail for reactive streams semantics, probably a link is best.

benjchristensen:
Agreed regarding the link, which is already there. It is only the cumulative versus reset point that needed to be clarified here I think.

PR #207, opened 2017-02-28
remove Stream State section

benjchristensen:
It is confusing and doesn't match terminology used elsewhere.

The CLOSED state is not used anywhere else, nor is REQUESTED state. 

This is related to https://github.com/ReactiveSocket/reactivesocket/issues/193 which sought clarification of this section.


benjchristensen:
If there is desire to not remove this section, how should it be re-written to address the confusion of discussing states never used elsewhere in the spec?

PR #214, opened 2017-03-01
remove (S)trict flag on SETUP

benjchristensen:
Unless we determine a reason to keep it, removing so it's not ambiguous: https://github.com/ReactiveSocket/reactivesocket/issues/209


fiorix:
+1 here, I don't really see the reason for this flag.

PR #215, opened 2017-03-01
Remove Metadata from CANCEL & ERROR

benjchristensen:
As per discussion in https://github.com/ReactiveSocket/reactivesocket/issues/201, remove Metadata from ERROR, LEASE, and CANCEL frames.

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -375,15 +375,18 @@ Frame Contents
     +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
     |                         Stream ID = 0                         |
     +-----------+-+-+---------------+-------------------------------+
-    |Frame Type |0|0|     Flags     |
+    |Frame Type |0|M|     Flags     |
     +-----------+-+-+---------------+-------------------------------+
     |0|                       Time-To-Live                          |
     +---------------------------------------------------------------+
     |0|                     Number of Requests                      |
     +---------------------------------------------------------------+
+                                Metadata
 ```
 
 * __Frame Type__: (6 bits = max value 63) 0x02 
+* __Flags__: (10 bits)
+     * (__M__)etadata: Metadata present
 * __Time-To-Live (TTL)__: (31 bits = max value 2^31-1 = 2,147,483,647) Unsigned 31-bit integer of Time (in milliseconds) for validity of LEASE from time of reception. Value MUST be > 0. 
 * __Number of Requests__: (31 bits = max value 2^31-1 = 2,147,483,647) Unsigned 31-bit integer of Number of Requests that may be sent until next LEASE. Value MUST be > 0. 
 
--- a/Protocol.md
+++ b/Protocol.md
@@ -318,16 +318,14 @@ Frame Contents
     +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
     |                           Stream ID                           |
     +-----------+-+-+---------------+-------------------------------+
-    |Frame Type |0|M|      Flags    |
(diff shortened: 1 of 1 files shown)


Protocol.md
```diff
@@ -318,16 +318,14 @@ Frame Contents
...
-                        Metadata & Error Data
+                               Error Data
 ```
 
 * __Frame Type__: (6 bits = max value 63) 0x0B
-* __Flags__: (10 bits)
-     * (__M__)etadata: Metadata present
 * __Error Code__: (32 bits = max value 2^31-1 = 2,147,483,647) Type of Error.
      * See list of valid Error Codes below.
 * __Error Data__: includes Payload describing error information. Error Data SHOULD be a UTF-8 encoded string. The string MUST NOT be null terminated.
```
yschimke: My reading of this is that Error data does not use either Metadata or Data mimetype.  Error data is always just a UTF-8 string.

Is that correct?

Protocol.md
```diff
@@ -377,18 +375,15 @@ Frame Contents
...
     +-----------+-+-+---------------+-------------------------------+
     |0|                       Time-To-Live                          |
     +---------------------------------------------------------------+
     |0|                     Number of Requests                      |
     +---------------------------------------------------------------+
-                                Metadata
 ```
 
 * __Frame Type__: (6 bits = max value 63) 0x02 
-* __Flags__: (10 bits)
```
yschimke: This removes the ability to use LEASE frames to coordinate higher level application SLAs.  i.e. this is the most likely reason I would use it.  Is that the intention?

Protocol.md
benjchristensen: Yes.

Protocol.md
benjchristensen: Sorry, I don't understand ... are you saying that a LEASE frame has a use case for including metadata? If so, what is it?

Protocol.md
yschimke: Specifically my question is how to handle the behaviour that thrift adopts where idl defined exceptions are strongly typed.

We remove the Metadata with agreed Mimetype and replace it with a String.  Now to encode something like a thrift error we need to map the idl error codes to the > 0x0300 error code.

More likely most error handling will become substring matching of known error strings

tmontgomery:
I am OK with these changes. I'll let @yschimke answer before approving. Or he can approve if he is OK.

Protocol.md
benjchristensen: If we did do something other than UTF-8, we'd need another MimeType, as we can't assume either Data or Metadata to be appropriate for this. Also, if the MimeType itself is the problem causing the error (bad serialization for example), then is there a fallback to UTF-8?

I would prefer maintaining the simplicity of UTF-8 for error codes so anything can read them even if they don't know the MimeType. 

For the Thrift IDL case, I would just turn each code into a UTF-8 string. For error cases I think the performance of doing so is not a big deal.

benjchristensen: @robertroeser @stevegury @NiteshKant what do you think?

benjchristensen: @fiorix ?

Protocol.md
stevegury: Yes, there are use cases where we use metadata, c.f. my response to https://github.com/ReactiveSocket/reactivesocket/issues/201
We rely on that in our java implementation, I would like to keep it that way.

benjchristensen: I have restored metadata for LEASE.

benjchristensen:
Based on @stevegury's feedback, I have restored metadata for LEASE.

Does it still make sense to remove metadata for CANCEL & ERROR?

robertroeser:
I could go either way with the error frame. Cancel frame is created internal, and the rs user never see them so I don't see a reason fro metadata there.

benjchristensen:
Great, then I'll merge this. Thank you Steve, Robert, Yuri, and Todd for your input.

PR #216, opened 2017-03-01
Request Fragmentation

benjchristensen:
Revision of how request fragmentation will be done with PAYLOAD frames as per discussion in https://github.com/ReactiveSocket/reactivesocket/issues/198


benjchristensen:
@tmontgomery you okay with these as well?

benjchristensen:
Oh, I now see @tmontgomery already approved these as well.

PR #220, opened 2017-03-02
stream id generation

benjchristensen:
Clarification derived from discussion in https://github.com/ReactiveSocket/reactivesocket/issues/208

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -210,6 +210,10 @@ A Stream ID MUST be locally unique for a Requester in a connection.
 Stream ID generation follows general guidelines for [HTTP/2](https://tools.ietf.org/html/rfc7540) with respect
 to odd/even values. In other words, a client MUST generate odd Stream IDs and a server MUST generate even Stream IDs.
 
+Stream IDs on the client MUST start at 1 and increment by 2 sequentially, such as 1, 3, 5, 7, etc.
+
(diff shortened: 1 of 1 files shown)


NiteshKant:
I am late to the party here but it seems a bit of a loose contract to me. What does it mean that the stream IDs must be sequential? Does it mean that the requester generates it sequentially or does it mean the responder receives it sequentially (since RS works on an ordered transport, it means the IDs were sent sequentially)?

If this is supposed to mean that the requester generates it sequentially, what value does it add?

If this is supposed to act as a validation by the responder to detect gaps (as suggested in the [original discussion](https://github.com/ReactiveSocket/reactivesocket/issues/208#issuecomment-283192722)) then actually there is a race here. Generating a stream ID sequentially does not guarantee sending them sequentially on the wire since multiple ID generations and writes can be interleaved.

benjchristensen:
It states in the spec with example:

> Stream IDs on the client MUST start at 1 and increment by 2 sequentially, such as 1, 3, 5, 7, etc.
> Stream IDs on the server MUST start at 2 and increment by 2 sequentially, such as 2, 4, 6, 8, etc.

Does that clarify it?

This is not to allow detection or prevent gaps. In my view the only benefit it gives is it significantly reduces the chance of collision on re-use when wrapping.

PR #225, opened 2017-03-06
Including Http2 as a supported trasport

mitermayer:
Making sure all references of supported transport include HTTP2 and also trimming down extra white spaces.

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -1,6 +1,6 @@
 ## Status
 
-This protocol is currently a draft for the final specifications. 
+This protocol is currently a draft for the final specifications.
 Current version of the protocol is __0.2__ (Major Version: 0, Minor Version: 2).
 
 ## Introduction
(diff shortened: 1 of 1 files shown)


Protocol.md
```diff
@@ -154,7 +154,7 @@ ReactiveSocket frames begin with a ReactiveSocket Frame Header. The general layo
 ```
 
 * __Stream ID__: (31 bits = max value 2^31-1 = 2,147,483,647) Unsigned 31-bit integer representing the stream Identifier for this frame or 0 to indicate the entire connection.
-  * Transport protocols that include demultiplexing, such as HTTP/2, MAY omit the Stream ID field if all parties agree. The means of negotiation and agreement is left to the transport protocol. 
+  * Transport protocols that include demultiplexing, such as HTTP/2, MAY omit the Stream ID field if all parties agree. The means of negotiation and agreement is left to the transport protocol.
```
yschimke: Is there a completely unambiguous and implementable public definition of what we mean by HTTP/2 support?

a) quasi-TCP over a single HTTP/2 stream
b) HTTP/2 extension (https://github.com/ReactiveSocket/reactivesocket/issues/84)
https://github.com/ReactiveSocket/reactivesocket/issues/84

This sentence seems to imply the latter implementation.

benjchristensen: When I updated the spec a couple weeks back to clarify HTTP/2, I adjusted to use "HTTP/2 stream" and not just "HTTP/2". We could add a section explaining what this is about. 

It remains somewhat ambiguous right now though.

benjchristensen:
@mitermayer Can you split out the HTTP/2 stuff from whitespace cleanup so the commit log is clear?

PR #229, opened 2017-03-07
Proposal for optional metadata length header

benjchristensen:
Proposed wording in case we choose to make Metadata Length header optional.

See https://github.com/ReactiveSocket/reactivesocket/issues/228


tmontgomery:
This looks good to me.

Issue #231, opened 2017-03-14
Metadata length question


josephsavona:
From https://github.com/ReactiveSocket/reactivesocket/blob/master/Protocol.md#metadata-optional-header:

> Metadata Length MUST be less than or equal to the Frame Length minus the length of the Frame Header. If Metadata Length is greater than this value, the entire frame MUST be ignored.

The only reason it could be less than the mentioned quantity is if there is payload data. Should this instead be "Metadata Length MUST be **equal** to the frame length minus the sum of (the length of the frame header + length of the frame payload, if present)"?

benjchristensen:
Read this a few times, and I think you're right that it could be made exact as you say. 

@tmontgomery What do you think of the wording change?

tmontgomery:
Yes, wording change should be good.

PR #232, opened 2017-03-16
Clarify metadata length value

josephsavona:
Per #231

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -177,7 +177,7 @@ ReactiveSocket implementations may provide their own validation at the metadata 
 
 Specific Frame Types MAY contain Metadata. If that Frame Type supports both Data and Metadata, the optional Metadata header MUST be included. This metadata header is between the Frame Header and any payload.
 
-Metadata Length MUST be equal to the Frame Length minus the sum of the length of the Frame Header and the length of the Frame Payload, if present. If Metadata Length is not equal to this value, the frame is invalid. The receiver MUST send an ERROR[CONNECTION_ERROR] frame and close the underlying transport connection on reception unless the frame's IGNORE flag is set.
+Metadata Length MUST be equal to the Frame Length minus the sum of the length of the Frame Header and the length of the Frame Payload, if present. If Metadata Length is not equal to this value, the frame is invalid and the receiver MUST send an ERROR[CONNECTION_ERROR] frame and close the underlying transport connection on reception unless the frame's IGNORE flag is set.
 
 On a frame with Data and Metadata:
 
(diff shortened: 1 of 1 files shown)


Protocol.md
```diff
@@ -177,8 +177,7 @@ ReactiveSocket implementations may provide their own validation at the metadata
 
 Specific Frame Types MAY contain Metadata. If that Frame Type supports both Data and Metadata, the optional Metadata header MUST be included. This metadata header is between the Frame Header and any payload.
 
-Metadata Length MUST be less than or equal to the Frame Length minus the length of the Frame Header.
-If Metadata Length is greater than this value, the entire frame MUST be ignored.
+Metadata Length MUST be equal to the Frame Length minus the sum of the length of the Frame Header and the length of the Frame Payload, if present. If Metadata Length is not equal to this value, the entire frame MUST be ignored.
```
yschimke: Isn't this tautological since there is not a separate length of the Frame Payload?

The length of the Frame Payload is derived from 

Frame length - Frame Header length - metadata length

josephsavona: Yes, but the payload length isn't encoded in the frame, so the only way to know where the metadata/payload boundary starts is to know the exact value of either metadata length, payload length, or payload starting offset.

yschimke:
I'm also confused what ignore means here. It can't mean pretend it never came. I think it means error the stream instead of delivering.

josephsavona:
I agree that propagating an error would seem reasonable/preferable to ignoring, since it would constitute invalid/corrupt data.

@benjchristensen Thoughts?

Protocol.md
josephsavona: cc @benjchristensen

benjchristensen:
Want to chat with Ondrej and confirm the existing implementation details.

Protocol.md
lehecka: I am fine with this clarification. I understand the spec the same way before and after.

lehecka:
I would definitely recommend changing the wording around "ignore frame" to "connection is closed with an error" or something like that. The c++ implementation framing code never "ignores" a frame when the boundaries are broken or if a field is outside of the range, etc. Debugging of "ignored" frames would be very hard.

benjchristensen:
I think I agree that we should error rather than ignore. 

@tmontgomery what do you think here?

tmontgomery:
Should error rather than ignore in this particular case as framing is probably off in a serious way. So, making it a hard error is warranted.

josephsavona:
cc @benjchristensen

PR #236, opened 2017-06-27
clear up resumption tracking

yschimke:
Clarify that position is byte-count of frames received, including fragmented frames.

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -736,13 +736,13 @@ RSocket resumption exists only for specific cases. It is not intended to be an �
 
 Resuming operation requires knowing the position of data reception of the previous connection. For this to be simplified, the underlying transport is assumed to support contiguous delivery of data on a per frame basis. In other words, partial frames are not delivered for processing nor are gaps allowed in the stream of frames sent by either the client or server. The current list of supported transports (TCP, WebSocket, and Aeron) all satisfy this requirement or can be made to do so in the case of TCP.
 
-As a Requester or Responder __sends__ REQUEST, CANCEL, or PAYLOAD frames, it maintains a __position__ of that frame within the connection in that direction. This is a 64-bit value that starts at 0. As a Requester or Responder __receives__ REQUEST, CANCEL, or PAYLOAD frames, it maintains an __implied position__ of that frame within the connection in that direction. This is also a 64-bit value that starts at 0.
+As a Requester or Responder __sends__ REQUEST_RESPONSE, REQUEST_FNF, REQUEST_STREAM, REQUEST_CHANNEL, REQUEST_N, CANCEL, ERROR, or PAYLOAD frames, it maintains a __position__ of that frame within the connection in that direction. This is a 64-bit value that starts at 0. As a Requester or Responder __receives__ those tracked frames, it maintains an __implied position__ of that frame within the connection in that direction. This is also a 64-bit value that starts at 0.  The positions are calculated based on the length of encoded frames without the frame length field after any fragmentation is applied.
 
 The reason this is “implied” is that the position is not included in each frame and is inferred simply by the message being sent/received on the connection in relation to previous frames.
 
 This position will be used to identify the location for resuming operation to begin.
 
-Frame types outside REQUEST, CANCEL, ERROR, and PAYLOAD do not have assigned (nor implied) positions.
+Frame types outside REQUEST(s), REQUEST_N, CANCEL, ERROR, and PAYLOAD do not have assigned (nor implied) positions.
 
 When a client sends a RESUME frame, it sends two implied positions: the last frame that was received from the server; the earliest frame position it still retains.  The server can make a determination on whether resumption is possible: have all frames past the client's last-received position been retained? and has the client retained all frames past the server's last-retained position.  If resumption is allowed to continue, the server sends a RESUME_OK frame, indicating its last-received position.
 


benjchristensen:
Should this be even more explicit that the position is the byte-count of frames received? And do we count a frame fragment, or only an assembled complete frame?

tmontgomery:
The intention was to count frame fragments. Reassembled frames becomes tricky. Suppose that resumption happens in the middle of a fragment, for example.

tmontgomery:
Being explicit about byte-count is probably good.

benjchristensen:
@yschimke if you can revise to be more clear about byte-count of all frames received (even if fragments), I would then be happy with this.

yschimke:
@benjchristensen define "all frames", even those not stored in the resume cache? Do you expect this to be the offset in the underlying bytes of a stream based transport?  Currently for rsocket-cpp it is 

```
if (isInteresting(frame)) 
  position += (encodedLength(frame))
```

yschimke:
@tmontgomery @benjchristensen @lexs 

Two potential issues:
1) critical frames are currently out of consideration for resumability, e.g. extension frame with stream id = 0
2) Can we clarify that REQUEST frame means, REQUEST_* including REQUEST_N but excluding METADATA_PUSH?

**The one thing that makes me nervous about a simple +1 strategy for rsocket v1.1, is unknown frames. They get dropped according to spec, but if one helpful implementor decides to count them or METADATA_PUSH then the protocol may not notice that the connection is out of sync.  Worse when out of sync, the work to repair the connection will make them more out of sync.**

This is based on:

> 9. All other received frames that are not accounted for in previous sections MUST be ignored.

What happens with extension frames? Should they be included in the count if not ignoreable? Should extension frames on the connection (stream id = 0) be counted?

> EXT (Extension) Frame (0x3F)
> (I)gnore: Can the frame be ignored if not understood?

So a protocol/implementation that can't decode a certain frame and storing frames in the Resume Cache may drop frames that are not ignoreable leading to data loss.

tmontgomery:
The spec should be explicit about the frames that constitute the tracking. This was the case with the spec last time I saw it. Extension frames are not covered by resumption as they are not, by definition, explicitly called out.

REQUEST frames are the frames that initiate a request. This is not REQUEST_N.

METADATA_PUSH can be added if there is a desire to have it included in a warm resume scenario. This is a decision for the users of resumption to make. If needed, can be added. If not, then leave it out.

yschimke:
without request_n it is possible for client and server to disagree about outstanding credits for a stream and starve a stream or channel.

Without extension, the protocol is broken for a extension frame on the connection stream being fatal.

This is the cpp code: https://github.com/rsocket/rsocket-cpp/blob/f154f0c1fde388c660c0f7ee3724bb954ad14c2b/src/internal/ResumeCache.cpp#L15

tmontgomery:
Extension frames have no real definition and are not required, so if they are broken... it is the protocol implementation of them that is broken. Any implementation that depends on extension frames is, by definition, not compliant with the spec. I know of no well designed protocol that would make extension frames mandatory to understand. It goes against the concept of extension and prevents forward compatibility.

To be clear, I did not say that REQUEST_N was not needed (didn't mean to imply it either, but I may accidentally have). I said REQUEST_N is not a request frame. i.e. was answering the specific question. So, my apologies.

We've discussed whether to cache REQUEST_N several times. We have never really come down on a solution. I believe it was left until more experience with running in operation could be done.

For REQUEST_N, yes, starvation can occur if Responder and Requester disagree on state. Which they very well may if REQUEST_N frames are not part of a warm resume. However, this is also a moving target. Even with making REQUEST_N part of a warm resume, the condition can exist that a warm resume can not be completed. In that case, individual streams would need to restart. Leaving the credit state unknown. And cold resume has even dirtier issues in this regard. Simply caching REQUEST_N or accounting for it (in the case of a cold resume) doesn't really solve the issue.

So, I've advocated a cleaner solution of assuming 1 credit on resume completion and allowing a single response to be sent upon resume completion instead of tracking REQUEST_N explicitly.

yschimke:
Should we use the ignore flag for extension frames to decide whether to track?

tmontgomery:
@yschimke Interesting idea. we could. But I would hesitate as it combines two different objectives.

The Ignore bit is to know whether understanding of the frame indicates a critical element of operation. Or can it be safely ignored if not understood. We use MAY in the spec. But it should probably be MUST since a compliant implementation is most likely to be conservative about it.

If NOT Ignore-able, then the frame indicates a cancelation of the connection. MAY indicate.

Whether to track for resume seems orthogonal to Ignore. But maybe adding a new bit for the extension header (I, M and maybe a new one T) that indicates it should be tracked for resume. That way while the content is still opaque (not understood), it could be tracked and the connection continue.

To be clear, this is what an extension frame flags (I and T) might be:

- (I)gnore: connection can continue. Frame can be safely ignored. No tracking for resume.
- (I)gnore + (T)rack:  connection can continue. Frame can be safely ignored. Tracking for resume.
- (T)rack: connection can only continue if frame understood. Tracking for resume.
- neither: connection can only continue if frame understood. No tracking for resume.

Thoughts?

yschimke:
@benjchristensen @lehecka  as this matches the c++ can I land this now? we still have @josephsavona PR which simplifies this for v1.1

PR #238, opened 2017-07-03
Clarify resumption position is frames not bytes

josephsavona:
Changes the spec to make it extra clear that the resumption implied positions are a count of frames and not bytes. Adding this since there was confusion and existing implementations diverge from the spec and use byte count.

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -737,7 +737,7 @@ RSocket resumption exists only for specific cases. It is not intended to be an �
 
 Resuming operation requires knowing the position of data reception of the previous connection. For this to be simplified, the underlying transport is assumed to support contiguous delivery of data on a per frame basis. In other words, partial frames are not delivered for processing nor are gaps allowed in the stream of frames sent by either the client or server. The current list of supported transports (TCP, WebSocket, and Aeron) all satisfy this requirement or can be made to do so in the case of TCP.
 
-As a Requester or Responder __sends__ REQUEST, CANCEL, or PAYLOAD frames, it maintains a __position__ of that frame within the connection in that direction. This is a 64-bit value that starts at 0. As a Requester or Responder __receives__ REQUEST, CANCEL, or PAYLOAD frames, it maintains an __implied position__ of that frame within the connection in that direction. This is also a 64-bit value that starts at 0.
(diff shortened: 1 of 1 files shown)


Protocol.md
```diff
@@ -737,15 +737,15 @@ RSocket resumption exists only for specific cases. It is not intended to be an 
...
 
-As a Requester or Responder __sends__ REQUEST, CANCEL, or PAYLOAD frames, it maintains a __position__ of that frame within the connection in that direction. This is a 64-bit value that starts at 0. As a Requester or Responder __receives__ REQUEST, CANCEL, or PAYLOAD frames, it maintains an __implied position__ of that frame within the connection in that direction. This is also a 64-bit value that starts at 0.
+As a Requester or Responder __sends__ REQUEST_FNF, REQUEST_RESPONSE, REQUEST_STREAM, REQUEST_CHANNEL, METADATA_PUSH, CANCEL, or PAYLOAD frames, it maintains a __position__ of that frame within the connection in that direction. This is a 64-bit value that starts at 0. As a Requester or Responder __receives__ REQUEST, CANCEL, or PAYLOAD frames, it maintains an __implied position__ of that frame within the connection in that direction. This is also a 64-bit value that starts at 0. Both the sender and receiver positions are the *number* of frames sent/received.
```
yschimke: I thought it was agreed that v1.0 remains byte lengths, and v1.1 changes to count.  Also we should be clear that it is a count of tracked frames.

yschimke: I also flattened the frame types in this PR - https://github.com/rsocket/rsocket/pull/236/files

josephsavona: How are we tracking changes to v1.0 vs v1.1? Is there a separate doc tracking the latter? If not should we make one?

josephsavona: Yup saw that. I wanted to make sure we didn't lose the proposed simplification for v1.1.

yschimke: milestones for now?  A branch once 1.0 is final. 

Not sure who can create those, but would be good to tag either issues or PRs.

cc @benjchristensen

yschimke: pinging @benjchristensen to create a 1.1 milestone for this PR

PR #254, opened 2018-10-04
Routing Extension Spec

nebhale:
This change adds an extension specification for routing that allows arbitrary
tags to be included by the requestor for the responder to use for message
routing.

> When two system are communicating via RSocket, there are often logical divisions in the messages that are sent from the requestor to the responder.  These logical divisions can often be implemenated by the responder as "routes" for messages to be sent to.  This extension specification provides an interoperable structure for metadata payloads to contain routing information.  It is designed such that an arbitrary collection of tags can be used by the responder to route messages and any individual tag (or all included tags) can be ignored.

### DIFF
--- a/Extensions/Routing.md
+++ b/Extensions/Routing.md
@@ -1,0 +1,30 @@
+# Routing Metadata Extension
+
+_This extension specification is currently incubating.  While incubating the version is 0._
+
+## Introduction
+When two system are communicating via RSocket, there are often logical divisions in the messages that are sent from the requester to the responder.  These logical divisions can often be implemented by the responder as "routes" for messages to be sent to.  This extension specification provides an interoperable structure for metadata payloads to contain routing information.  It is designed such that an arbitrary collection of tags (strings) can be used by the responder to route messages and any individual tag (or all included tags) can be ignored.
+
+## Metadata Payload
+This metadata type is intended to be used per stream, and not per connection nor individual payloads and as such it **MUST** only be used in frame types used to initiate interactions.  This includes [`REQUEST_FNF`][rf], [`REQUEST_RESPONSE`][rr], [`REQUEST_STREAM`][rs], and [`REQUEST_CHANNEL`][rc].  The Metadata MIME Type is `message/x.rsocket.routing.v0`.
+
+[rc]: ../Protocol.md#frame-request-channel
+[rf]: ../Protocol.md#frame-fnf
+[rr]: ../Protocol.md#frame-request-response
+[rs]: ../Protocol.md#frame-request-stream
+
+### Metadata Contents
+```
+     0                   1                   2                   3
+     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+    |  Tag Length   |              Tag                             ...
+    +---------------+-----------------------------------------------+
+    |  Tag Length   |              Tag                             ...
+    +---------------+-----------------------------------------------+
+                                   ...
+```
+
+* **Tag Payload**: Any number of complete tag payloads.
+  * **Tag Length**: Tag Length in bytes.
+  * **Tag**:  Token used for routing.  The string MUST NOT be null terminated.  Examples include URI-style routes (`/person`, `/address`), or artibrary metadata (`ios-client`, `android-client`).
--- a/Extensions/CompositeMetadata.md
+++ b/Extensions/CompositeMetadata.md
@@ -46,4 +46,4 @@ This metadata type is intended to be used per stream, and not per connection nor
 [rf]: https://tools.ietf.org/html/rfc2045
 [s]:  http://www.iana.org/assignments/media-type-structured-suffix/media-type-structured-suffix.xml
 [tz]: Tracing-Zipkin.md
-[wk]: Well-Known-MIME-Types.md
+[wk]: WellKnownMimeTypes.md
--- a/Extensions/WellKnownMimeTypes.md
+++ b/Extensions/WellKnownMimeTypes.md
@@ -48,4 +48,5 @@ All well-known MIME types assume UTF-8 character encoding wherever a character s
 | `video/VP8` | `37`
 | |
 | `message/x.rsocket.tracing-zipkin.v0` | `125`
+| `message/x.rsocket.routing.v0` | `126`
 | `message/x.rsocket.composite-metadata.v0` | `127`
--- a/README.md
+++ b/README.md
@@ -15,8 +15,9 @@ Artifacts include:
 
 - [RSocket Protocol](Protocol.md)
 - Protocol Extensions
-  - [Well-known MIME Types](Extensions/Well-Known-MIME-Types.md)
+  - [Well-known MIME Types](Extensions/WellKnownMimeTypes.md)
   - [Composite Metdata](Extensions/CompositeMetadata.md)
+  - [Routing](Extensions/Routing.md)
   - [Tracing (Zipkin)](Extensions/Tracing-Zipkin.md)
 - [Motivation for creating this protocol](Motivations.md)
 - [Frequently Asked Questions](FAQ.md)


nebhale:
@nebhale
> I've put in the distributed tracing since we discussed it, but I'm not sure how I feel about it. It feels like we get 10 shots at "semantic" things we want to put in (the size of the flags) rather than a generic way that is infinitely expandable. But then, I'm beginning to think that a characteristic of a good design is that it is bounded in this way.

nebhale:
@robertroeser 
> Should there be a payload encoding MIME_TYPE as well? eg - application/x.spring+json or application/x.proteus+protobuf

nebhale:
@nebhale 
> @robertroeser I’d wondered the same thing. In this case, we’re having a PAYLOAD-specific metadata MIME Type because we’ve used the one in the SETUP frame. We won’t have used the data MIME Type in the SETUP frame yet so it’s still in play. However, one of the advantages of this metadata format is that each PAYLOAD can have it’s own metadata MIME Type and maybe getting that same feature for data is useful as well. What would you put in for the data MIME Type in the SETUP frame?

nebhale:
@nebhale 
> I’m also interested if anyone has an opinion on whether the distributed tracing bit needs a MIME Type. I don’t know enough about the subject to determine if there are multiple competing binary encodings there.

nebhale:
@rdegnan 

> I found this describing the Zipkin tracing format, it doesn't have a MIME Type, but it does have more than one token -- we'll probably need at least the following:
>
> Trace Id: The overall 64 or 128-bit ID of the trace. Every span in a trace shares this ID.
> Span Id: The ID for a particular span. This may or may not be the same as the trace id.
> Parent Id: This is an optional ID that will only be present on child spans. That is the span without a parent id is considered the root of the trace.
>
> Plus maybe a boolean flag to tell it whether or not to record the trace.

nebhale:
@nebhale 
> @rdegnan we’d have to support the debug flag as well there, so two loose bytes at the end. My bigger question (and one I’ll run past the Spring Cloud Sleuth and Zipkin teams at Pivotal today) is whether the Zipkin wire-format is the only one we need to care about. I’m wonder if there’s a competing standard (OpenTracing?) out there.

nebhale:
@nebhale 
> Updated with optional tracing payload. I think this is done, pending comments.

nebhale:
@rdegnan 

> @nebhale This looks great. As Robert and I started to try this out last week one thing we discussed was instead of having a Route Token, just having a full-blown Routing Metadata with its own MIME Type, which would allow different routing systems to coexist and be delegated to depending on which MIME types they handle. So something like this:
>
```
    |            Version            |T|P|S|D|   Flags   |
    +-------------------------------+-+-+-+-+-----------+-----------+
    |                                                               |
    +                                                               +
    |                                                               |
    +                           Trace ID                            +
    |                                                               |
    +                                                               +
    |                                                               |
    |                                                               |
    +                           Span ID                             +
    |                                                               |
    |                                                               |
    +                        Parent Span ID                         +
    |                                                               |
    +-+-+-----------------------------------------------------------+
    |S|D|
    +-+-+-----------+-----------------------------------------------+
    |  MIME Length  |   Routing Metadata Encoding MIME Type        ...
    +---------------+-----------------------------------------------+
    |  MIME Length  |   Metadata Encoding MIME Type                ...
    +---------------+-----------------------------------------------+
    |  MIME Length  |   Data Encoding MIME Type                    ...
    +---------------+-------------------------------+---------------+
    |          Routing Metadata Length              |
    +-----------------------------------------------+---------------+
    |                   Routing Metadata Payload                   ...
                          Metadata Payload
```
>
> Any cons you see to this approach?

nebhale:
@rdegnan Picking up, it does seem more flexible, and my one concern (and [others](
https://twitter.com/nebhale/status/1045111010254901248)) is about overhead.  I defer to your expertise around this so if you think the flexibility is worth it, then I'm on board.

robertroeser:
I'd prefer tracing was an optional variable length field and then people could put in whatever they wanted and not be tied to something else.

Routing.md
```diff
@@ -0,0 +1,59 @@
+# Routing Meatdata Payload Extension
```
yschimke: Meatdata *he he*

Routing.md
```diff
@@ -0,0 +1,59 @@
...
+    +---------------------------------------------------------------+
+    |                                                               |
+    +                           Span ID                             +
+    |                                                               |
+    +---------------------------------------------------------------+
+    |                                                               |
+    +                        Parent Span ID                         +
+    |                                                               |
+    +-+-+-----------------------------------------------------------+
+    |S|D|
```
yschimke: Why are these two bits not encoded in the flags? I assume I’m missing something.

Routing.md
```diff
@@ -0,0 +1,59 @@
...
+    +-+-+-----------+-----------------------------------------------+
+    |  MIME Length  |   Metadata Encoding MIME Type                ...
+    +---------------+-----------------------------------------------+
+    |  MIME Length  |   Data Encoding MIME Type                    ...
+    +---------------+-----------------------------------------------+
+                          Metadata Payload
+```
+
+* **Version**: (16 bits = max value 65,535) Unsigned 16-bit integer of version number of the routing payload
+* **Flags**: (10 bits)
```
yschimke: Isn’t byte aligning easier? E.g. padding out to byte and not making things optional at bit level

yschimke: Even just for sanity of people reading the hex stream.

nebhale:
@yschimke this matches @robertroeser's [previous statement](
https://github.com/rsocket/rsocket/pull/254#issuecomment-427135698).  It's been a long enough time since we did the initial straw man that I can't remember why we chose to do that.  I think it was just an acceptance that all the tracing formats are constant now?

The one thing I would caution about is that as everything becomes flexible and self-defining via MIME types, we don't end up specing very much and depending on a priori knowledge for interop.

fiorix:
Hey, noob here. I just read the gist in the first message, and followed this thread, and I'm failing to understand the use case for this at the protocol level, as an extension.

Isn't the protocol flexible enough for users to implement routing in the application layer? e.g. if their metadata is some kind of dict/map, like in http, proxies and routers work just fine.

Could someone please clarify?

robertroeser:
@fiorix it is flexible- but there is no common routing like http URLs so everyone keeps re-inventing the wheel - I have see at least 3 different versions so far - I'm sure FB has more internally. There is no way to tell what "their data" is without something common - if I build a router how do I know where to route your data if I can't read it? This tries to create to a suggested format for people to use on top of RSocket so we don't have multiple different users each with a unique routing format.

Routing.md
```diff
@@ -0,0 +1,59 @@
...
+  * (**T**)racing: Payload contains distributed tracing tokens
```
smaldini: Am i correct to note that will provide a one time payload for tracing all messages in request/stream or channel scenarios ? Put differently, is this going to trace request/stream and channel entirely vs per-message basis.

yschimke:
Yeah some some of standard default would be useful if you might ever say “supported in envoy and linkerd” etc. Facebook routing was off well known keys in json metadata.

yschimke:
@nebhale maybe just draw it as a flexible byte range in the diagram. And then in this spec also say that application/opentracing is 8 packed byes (4-2-2)

Routing.md
nebhale: This payload would be part of the metadata of each `PAYLOAD` frame.  So each `PAYLOAD` _could_ have it's own trace ID within a parent span if desired.  I'm not sure what we want to happen here, so it might be worthwhile to explore and decide if we want to specify something.

Routing.md
```diff
@@ -0,0 +1,59 @@
+# Routing Metadata Payload Extension
+
+## Introduction
+Specify a metadata payload for [RSocket](https://rsocket.io) that contains both information about where to route a [`PAYLOAD` frame](https://github.com/rsocket/rsocket/blob/master/Protocol.md#frame-payload) as well as metadata for the frame itself.
```
nebhale: Improve documentation to describe where this goes in a flow.

Routing.md
nebhale: Splitting out tracing into a separate extension spec.

Routing.md
nebhale: Moving out into a tracing spec.

Extensions/Routing.md
```diff
@@ -0,0 +1,30 @@
+# Routing Metadata Extension
+
+_This extension specification is currently incubating.  While incubating the version is 0._
+
+## Introduction
+When two system are communicating via RSocket, there are often logical divisions in the messages that are sent from the requestor to the responder.  These logical divisions can often be implemenated by the responder as "routes" for messages to be sent to.  This extension specification provides an interoperable structure for metadata payloads to contain routing information.  It is designed such that an arbitrary collection of tags can be used by the responder to route messages and any individual tag (or all included tags) can be ignored.
```
yschimke:
```suggestion
When two system are communicating via RSocket, there are often logical divisions in the messages that are sent from the requestor to the responder.  These logical divisions can often be implemented by the responder as "routes" for messages to be sent to.  This extension specification provides an interoperable structure for metadata payloads to contain routing information.  It is designed such that an arbitrary collection of tags can be used by the responder to route messages and any individual tag (or all included tags) can be ignored.
```

yschimke:
Are these tags a list of arbitrary opaque bytes, that client and server should negotiate ahead of time?

Are there any examples, or a sample mapping from an existing similar scheme? e.g. IP addresses?

nebhale:
@yschimke On the call we come to a loose agreement that tags would be opaque strings. (The docs mention that the string shouldn't be NULL terminated, but isn't explicit about them being strings in the first place, so I'll update that.)  The plan was that the requestor would attach a set of tags and the responder could choose to use those tags in any way that it saw fit for routing.  If that meant using all of them to establish a route, ignoring some that weren't important, etc., everything was possible.

szihai:
I would suggest different router implementations should be able to route each other's messages, at least not dropping them.

spencergibb:
Ping @adriancole re: tracing

PR #255, opened 2018-10-18
Composite Metadata Extension Spec

nebhale:
This change adds an extension specific for composite metadata that allows discrete metadata types to be included in a single metadata payload.

> There are a number of situations where an arbitrary collection of discrete metadata types should be attached to frame. For example, a request frame may want to include both routing metadata as well as tracing metadata. This extension specification provides an interoperable structure for metadadata payloads to contain multiple discrete metadata types. It is designed such that if a consumer of the metadata is unaware of a particular type, it can be safely skipped and the next one read.

### DIFF
--- a/Extensions/CompositeMetadata.md
+++ b/Extensions/CompositeMetadata.md
@@ -1,0 +1,49 @@
+# Composite Metadata Extension
+
+_This extension specification is currently incubating.  While incubating the version is 0._
+
+## Introduction
+There are a number of situations where an arbitrary collection of discrete metadata types should be attached to frame.  For example, a request frame may want to include both routing metadata as well as tracing metadata.  This extension specification provides an interoperable structure for metadadata payloads to contain multiple discrete metadata types.  It is designed such that if a consumer of the metadata is unaware of a particular type, it can be safely skipped and the next one read.
+
+## Metadata Payload
+This metadata type is intended to be used per stream, and not per connection nor individual payloads and as such it **MUST** only be used in frame types used to initiate interactions.  This includes [`REQUEST_FNF`][rf], [`REQUEST_RESPONSE`][rr], [`REQUEST_STREAM`][rs], and [`REQUEST_CHANNEL`][rc].  Multiple metadata payloads with the same MIME type are allowed.  The order of metadata payloads MUST be preserved when presented to responders.  The [`SETUP` Frame][s] Metadata MIME Type is `message/x.rsocket.composite-metadata.v0`.
+
+[rc]: ../Protocol.md#frame-request-channel
+[rf]: ../Protocol.md#frame-fnf
+[rr]: ../Protocol.md#frame-request-response
+[rs]: ../Protocol.md#frame-request-stream
+[s]:  ../Protocol.md#frame-setup
+
+### Metadata Contents
+```
+     0                   1                   2                   3
+     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+    |M| MIME ID/Len |   Metadata Encoding MIME Type                ...
+    +---------------+---------------+---------------+---------------+
+    |              Metadata Length                  |
+    +-----------------------------------------------+---------------+
+    |                     Metadata Payload                         ...
+    +---------------+-----------------------------------------------+
+    |M| MIME ID/Len |   Metadata Encoding MIME Type                ...
+    +---------------+-------------------------------+---------------+
+    |              Metadata Length                  |
+    +-----------------------------------------------+---------------+
+    |                     Metadata Payload                         ...
+    +---------------------------------------------------------------+
+                                   ...
+```
+
+* **Metadata Payload**: Any number of complete metadata payloads.
+  * (**M**)etadata Type: Metadata type is a well known value represented by a unique integer.
+  * **MIME ID/Length**: (7 bits = max value 2^7 = 128) Unsigned 7-bit integer.  If M flag is set, indicates a [Well-known MIME Type ID][wk].  If M flag is not set, indicates the encoding MIME Type Length in bytes.
+  * **Metadata Encoding MIME Type**: MIME Type for encoding of Metadata. This SHOULD be a US-ASCII string that includes the [Internet media type](https://en.wikipedia.org/wiki/Internet_media_type) specified in [RFC 2045][rf].  Many are registered with [IANA][ia] and others such as [Routing][r] and [Tracing (Zipkin)][tz] are not.  [Suffix][s] rules MAY be used for handling layout.  The string MUST NOT be null terminated.  (Not present if M flag is set)
+  * **Metadata Length**: (24 bits = max value 16,777,215) Unsigned 24-bit integer of Metadata Length in bytes.
+  * **Metadata Payload**: User configured metadata encoded as defined by the Metadata Encoding MIME Type.
+
+[ia]: https://www.iana.org/assignments/media-types/media-types.xhtml
+[r]:  Routing.md
+[rf]: https://tools.ietf.org/html/rfc2045
+[s]:  http://www.iana.org/assignments/media-type-structured-suffix/media-type-structured-suffix.xml
+[tz]: Tracing-Zipkin.md
+[wk]: Well-Known-MIME-Types.md
--- a/Extensions/Well-Known-MIME-Types.md
+++ b/Extensions/Well-Known-MIME-Types.md
@@ -46,3 +46,5 @@ All well-known MIME types assume UTF-8 character encoding wherever a character s
 | `video/H264` | `35`
 | `video/H265` | `36`
 | `video/VP8` | `37`
+| |
+| `message/x.rsocket.composite-metadata.v0` | `127`
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ Artifacts include:
 - [RSocket Protocol](Protocol.md)
 - Protocol Extensions
   - [Well-known MIME Types](Extensions/Well-Known-MIME-Types.md)
+  - [Composite Metdata](Extensions/CompositeMetadata.md)
 - [Motivation for creating this protocol](Motivations.md)
 - [Frequently Asked Questions](FAQ.md)
 


robertroeser:
I think this looks good. One thought I have is that we reserve the first bit in the mime-type length. If the bit is set instead of sending a mime type string it’s a well define mime-type integer. Ie 1 is tracing, 2 is routing etc.

Extensions/CompositeMetadata.md
```diff
@@ -0,0 +1,45 @@
...
+    |                     Metadata Payload                         ...
+    +---------------------------------------------------------------+
+                                   ...
+```
+
+* **Version**: (16 bits = max value 65,535) Unsigned 16-bit integer of version number of the composite metadata payload.
+* **Metadata Payload**: Any number of complete metadata payloads.
+  * **MIME Length**: Encoding MIME Type Length in bytes.
+  * **Metadata Encoding MIME Type**: MIME Type for encoding of Data and Metadata. This SHOULD be a US-ASCII string that includes the [Internet media type](https://en.wikipedia.org/wiki/Internet_media_type) specified in [RFC 2045](https://tools.ietf.org/html/rfc2045). Many are registered with [IANA](https://www.iana.org/assignments/media-types/media-types.xhtml) such as [CBOR](https://www.iana.org/assignments/media-types/application/cbor). [Suffix](http://www.iana.org/assignments/media-type-structured-suffix/media-type-structured-suffix.xml) rules MAY be used for handling layout. For example, `application/x.netflix+cbor` or `application/x.reactivesocket+cbor` or `application/x.netflix+json`. The string MUST NOT be null terminated.
```
yschimke: I think examples like the zipkin types are better for explaining how it will be used.

Extensions/CompositeMetadata.md
```diff
@@ -0,0 +1,45 @@
...
+    |  MIME Length  |   Metadata Encoding MIME Type                ...
+    +---------------+-------------------------------+---------------+
+    |              Metadata Length                  |
+    +-----------------------------------------------+---------------+
+    |                     Metadata Payload                         ...
+    +---------------------------------------------------------------+
+                                   ...
+```
+
+* **Version**: (16 bits = max value 65,535) Unsigned 16-bit integer of version number of the composite metadata payload.
```
yschimke: 2 bytes here seems like overkill.  Could either be part of the top level metadata mimetype when needed to be incremented, or possibly a single byte with space for flags?

Extensions/CompositeMetadata.md
```diff
@@ -0,0 +1,45 @@
+# Composite Metadata Extension
+
+_This extension specification is currently incubating._
+
+## Introduction
+There are a number of situations where an arbitrary collection of discrete metadata types should be attached to frame.  For example, a request frame may want to include both routing metadata as well as tracing metadata.  This extension specification provides an interoperable structure for metadadata payloads to contain multiple discrete metadata types.  It is designed such that if a consumer of the metadata is unaware of a particular type, it can be safely skipped and the next one read.
```
yschimke: I wonder whether safely skipping an unknown header should be a shared flag?  Should it be possible to require a responder to understand the metadata to proceed?

Extensions/CompositeMetadata.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Metadata Payload
+This metadata type **MUST** only be used in frame types used to initiate streams.  This includes [`REQUEST_FNF`][rf], [`REQUEST_RESPONSE`][rr], [`REQUEST_STREAM`][rs], and [`REQUEST_CHANNEL`][rc].  The [`SETUP` Frame][s] Metadata MIME Type is `message/x.rsocket.composite-metadata`.
```
yschimke: Nit: “initiate streams” doesn’t sit well with me for rf and rr.

Extensions/CompositeMetadata.md
yschimke: Is there guidance on using this for single metadata blocks? MAY be used?

Extensions/CompositeMetadata.md
robertroeser: Yeah when I mentioned we should version this I didn’t about the mime-type for metadata - we could probably version this there and make that part of the spec.

Extensions/CompositeMetadata.md
robertroeser: Maybe? Like if you sent a frame with routing data you could for it throw an exception if it didn’t understand?

yschimke:
This looks good, when I saw the zipkin spec first, my first impression was something like this would help :)

robertroeser:
what happens if you include multiple items with the same mime-type - ie two application/x.reactivesocket+cbor?

Or do we only allow one of each mine type. I think if we allow more than one we'll need to have some kind of name or label to tell them apparent.

Extensions/CompositeMetadata.md
nebhale: That isn't a required semantic for any other use of metadata in the spec though.  If you put a metadata MIME-type in ` SETUP` frame, there's no _mandate_ that the server fail if it isn't understood.

nebhale: If you're using a single metadata type, you would just write the metadata without using this type at all, right?  I'm not sure the spec needs to define, when _not_ to use it?

Extensions/CompositeMetadata.md
nebhale: I'll change it to "...to initiate interactions."

As to not `SETUP` and not `PAYLOAD`, on the call that resulted in this PR we discussed the idea that the proper level of granularity for this kind of extension was per-stream.  You'd set tracing metadata, or routing metadata on a per-stream basis and it didn't make sense for either connection-level or individual payloads. The plan was to take these three specs into incubation with this restriction and start using them to see if that hypothesis held.  If it didn't we broaden the spec to include other granularity.

Extensions/CompositeMetadata.md
nebhale: So is the consensus that we should have a MIME type of `message/x.rsocket.composite-metadata.v[\d]+` and remove the payload version?

Extensions/CompositeMetadata.md
nebhale: More important useful than consistency with the main spec?  That's my main reasoning for this text.

Extensions/CompositeMetadata.md
yschimke: OK, just raising.  I tend to find every potential ambiguity like this ends up with an implementor making different choices e.g. you receive this block with 0 metadata payloads instead of changing the top level flag :)

But TIOLI

yschimke:
> what happens if you include multiple items with the same mime-type - ie two application/x.reactivesocket+cbor?
> 
> Or do we only allow one of each mine type. I think if we allow more than one we'll need to have some kind of name or label to tell them apparent.

Also ordering, should it be maintained, e.g. when set on Java client (with or without conflicting dupes), will a JS server see blocks in same order?

Extensions/CompositeMetadata.md
nebhale: Given that we're still in the situation where the metadata MIME type for a connection is defined as part of `SETUP` I think we're going to have 0 and 1 inner metadatas when payloads are not homogenous.

@rdegnan, @robertroeser, @smaldini, or @stevegury any tie-breaking opinions?  Should we make some statement about using composite metadata in situations where it is unnecessary, but otherwise harmless (other than wasted bytes for a single)?

Extensions/CompositeMetadata.md
nebhale: OK, I've taken another crack at this mentioning routing and tracing.  Happy to keep iterating on the wording here.

nebhale:
My vote is for allowing multiple of the same MIME type, and preserving ordering.  I'll write up some text on it.

Extensions/CompositeMetadata.md
robertroeser: One difference between this and setup is that setup comes one time. Heres a list of well know mime types:

http://hul.harvard.edu/ois/////systems/wax/wax-public-help/mimetypes.htm

In addition to these we can add our own and then base the know list of the version. 

I would be ok with purposing this for rsocket setup as well.

Extensions/CompositeMetadata.md
```diff
@@ -0,0 +1,43 @@
...
+    |              Metadata Length                  |
+    +-----------------------------------------------+---------------+
+    |                     Metadata Payload                         ...
+    +---------------------------------------------------------------+
+                                   ...
+```
+
+* **Metadata Payload**: Any number of complete metadata payloads.
+  * (**M**)etadata Type: Metadata type is a well known value represented by a unique integer.
+  * **MIME Length**: (7 bits = max value 2^7 = 128) Unsigned 7-bit integer of the encoding MIME Type Length in bytes.  (Assumed to be 1 if M flag is set)
```
yschimke: If M is set, would the "Metadata Encoding MIME Type" be of length 0? and the rest of "MIME Length" is the value?  The "Assumed to be 1 if M flag is set" confuses me.

nebhale: @yschimke I would expect the MIME type to implicitly be an `byte`, and therefore the length is 1 byte.  From there the MIME type itself is just a unsigned byte value instead of a string.

Extensions/CompositeMetadata.md
nebhale: @robertroeser I'm not sure I'm following you here.  Are you stating that you'd like to support the entire IETF MIME type registry (as a starting point) as well-known MIME types?  That any of them could have an integer value equivalent?  I'd assumed that most types would be done with strings but spec'd extensions could have a well-known id since they're close to the project.

Extensions/CompositeMetadata.md
yschimke: OK, I had assumed this was all going to be packed in a single byte with 127 magic values, instead of two bytes (-1 flag + 1 byte length, then the 1 byte with 255 values)

Extensions/CompositeMetadata.md
yschimke: I liked the explicit mapping of types related to rsocket.  But suggest we extend it to support some well known ones likely to be used e.g. application/octet-stream, json, xml, etc + our zipkin format.

robertroeser: @nebhale @yschimke  If we are going to support only certain types, then we never send a mime-type string and only send binary. If we are going to allow a string and arbitrary types then I suggest we create an integer for all the well know types in addition to the ones we support so we aren't send tons of extra data.

Extensions/CompositeMetadata.md
nebhale: Well, given @robertroeser's desire to support all of the well known MIME types, 127 isn't going to be enough options.

Extensions/CompositeMetadata.md
nebhale: I have some concern about trying to assign an integer to every well-known MIME types as that list is fluid.  It grows, so we'd have to update our integer assignments regularly to support that growth.  Not sure how we'd version our well known-to-integer mapping either.

Extensions/CompositeMetadata.md
yschimke: I don't consider numbering all well known mime types to be viable given its a mutable set.  I hope you decide to just list ones that are widely applicable.

Extensions/CompositeMetadata.md
robertroeser: > I have some concern about trying to assign an integer to every well-known MIME types as that list is fluid.
By well know mime-types - I mean the commonly used ones - that's the list I included as a starting place, and come up with a list.

> Not sure how we'd version our well known-to-integer mapping either.
You create a static table and then you increment an integer as you add more types that you want to support. If its not on the table you send a string.

Extensions/CompositeMetadata.md
nebhale: @yschimke Another rev submitted.  Length is not included if M flag is set.

Extensions/CompositeMetadata.md
```diff
@@ -0,0 +1,49 @@
...
+
+* **Metadata Payload**: Any number of complete metadata payloads.
+  * (**M**)etadata Type: Metadata type is a well known value represented by a unique integer.
+  * **MIME ID/Length**: (7 bits = max value 2^7 = 128) Unsigned 7-bit integer.  If M flag is set, indicates a [Well-known MIME Type ID][wk].  If M flag is not set, indicates the encoding MIME Type Length in bytes.
+  * **Metadata Encoding MIME Type**: MIME Type for encoding of Metadata. This SHOULD be a US-ASCII string that includes the [Internet media type](https://en.wikipedia.org/wiki/Internet_media_type) specified in [RFC 2045][rf].  Many are registered with [IANA][ia] and others such as [Routing][r] and [Tracing (Zipkin)][tz] are not.  [Suffix][s] rules MAY be used for handling layout.  The string MUST NOT be null terminated.  (Not present if M flag is not set)
+  * **Metadata Length**: (24 bits = max value 16,777,215) Unsigned 24-bit integer of Metadata Length in bytes.
+  * **Metadata Payload**: User configured metadata encoded as defined by the Metadata Encoding MIME Type.
+
+[ia]: https://www.iana.org/assignments/media-types/media-types.xhtml
+[r]:  Routing.md
```
rstoyanchev: I don't see `Routing.md` in this branch.

rstoyanchev:
> on the call that resulted in this PR we discussed the idea that the proper level of granularity for this kind of extension was per-stream. You'd set tracing metadata, or routing metadata on a per-stream basis and it didn't make sense for either connection-level or individual payloads.

This is useful information. Perhaps it could be mentioned somewhere? Something like "this metadata type is intended to be used per stream, and not per connection nor for individual payloads, and as such it can only be used in frames that...".

For indexing, mime types can have parameters, e.g. `charset` is a very common one. One option would be to index a mime type multiple times per encoding, at least for the most common types. Alternatively do the opposite and explicitly state that all text based content (e.g. `"text/*"`, `"application/json"`, etc.) is UTF-8 by default. For any other encoding you'd have to use the charset parameter.

This extension does a good job of sub-dividing metadata into multiple independent areas. However it doesn't do anything to define a common way to access to things that are apply to everyone (or maybe that's being defined elsewhere?) For example, routing should be opaque but there shouldn't there be a simple (default) way for anyone to look up routing information? Another example is per-stream data encoding mime type, i.e. overriding the one negotiated in the SETUP.

Extensions/CompositeMetadata.md
nebhale: Yep, it goes in after this.  Chicken and egg, so it'll be a broken link for a while.

nebhale:
@rstoyanchev Cleaned up the wording to be more explicit about stream-based.

@yschimke Fixed typo.

@robertroeser I'd like you or @rdegnan as the third review approver if you're ready.

PR #256, opened 2018-10-18
Tracing (Zipkin)

nebhale:
This change adds an extension specification for distributed tracing in the Zipkin format.

> Observability and tracing are key requirements for robust and reliable applications.  When using distributed applications connected with RSocket, it's important to propagate metadata about the current logical operations throughout the entire system.  One of the most popular systems for doing this kind of tracing is [Zipkin][z].  This extension specification provides an interoperatble structure for Zipkin metadata payloads to contain tracing information.  It is designed such that systems can efficently communicate span and trace information to a Zipkin server and propagate that information throughout a distributed system.

[z]: https://zipkin.io

### DIFF
--- a/Extensions/Tracing-Zipkin.md
+++ b/Extensions/Tracing-Zipkin.md
@@ -1,0 +1,54 @@
+# Tracing (Zipkin) Metadata Extension
+
+_This extension specification is currently incubating.  While incubating the version is 0._
+
+## Introduction
+Observability and tracing are key requirements for robust and reliable applications.  When using distributed applications connected with RSocket, it's important to propagate metadata about the current logical operations throughout the entire system.  One of the most popular systems for doing this kind of tracing is [Zipkin][z].  This extension specification provides an interoperable structure for Zipkin metadata payloads to contain tracing information.  It is designed such that systems can efficently communicate span and trace information to a Zipkin server and propagate that information throughout a distributed system.
+
+[z]: https://zipkin.io
+
+## Metadata Payload
+This metadata type is intended to be used per stream, and not per connection nor individual payloads and as such it  **MUST** only be used in frame types used to initiate interactions and payloads.  This includes [`REQUEST_FNF`][rf], [`REQUEST_RESPONSE`][rr], [`REQUEST_STREAM`][rs], [`REQUEST_CHANNEL`][rc], and [`PAYLOAD`][p].  The Metadata MIME Type is `message/x.rsocket.tracing-zipkin.v0`.
+
+[p]:  ../Protocol.md#frame-payload
+[rc]: ../Protocol.md#frame-request-channel
+[rf]: ../Protocol.md#frame-fnf
+[rr]: ../Protocol.md#frame-request-response
+[rs]: ../Protocol.md#frame-request-stream
+
+### Metadata Contents
+```
+     0                   1                   2                   3
+     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+    |P|S  |D  |     |
+    +-+---+---+-----------------------------------------------------+
+    |                                                               |
+    +                                                               +
+    |                                                               |
+    +                           Trace ID                            +
+    |                                                               |
+    +                                                               +
+    |                                                               |
+    +---------------------------------------------------------------+
+    |                                                               |
+    +                           Span ID                             +
+    |                                                               |
+    +---------------------------------------------------------------+
+    |                                                               |
+    +                        Parent Span ID                         +
+    |                                                               |
+    +---------------------------------------------------------------+
+```
+
+* **Flags**: (8 bits)
+  * (**P**)arent Span Id: Tracing payload contains a parent span id.
+  * (**S**)ampling Decision (2 bits): Tracing payload contains a sample value. (Exclusive with D flag.)
+    * First bit indicates if a sampling decision is present
+    * Second bit indicates that the span should be reported to the tracing system. (Not present if first bit is not set)
+  * (**D**)ebug (2 bits): Tracing payload contains a debug value. (Exclusive with S flag.)
+    * First bit indicates if a debug decision is present
+    * Second bit indicates that the trace should be reported to the tracing system and also override any collection-tier sampling policy. Debug implies an accept sampling decision. (Not present if the first bit is not set)
+* **Trace ID**: (128 bits) Unsigned 128-bit integer ID of the trace. Every span in a trace shares this ID.
+* **Span ID**: (64 bits) Unsigned 64-bit integer ID for a particular span. This may or may not be the same as the trace id.
+* **Parent Span ID**: (64 bits) Unsigned 64-bit integer ID for a particular parent span.  This is an optional ID that will only be present on child spans. That is the span without a parent id is considered the root of the trace. (Not present if P flag is not set)
--- a/Extensions/Well-Known-MIME-Types.md
+++ b/Extensions/Well-Known-MIME-Types.md
@@ -47,4 +47,5 @@ All well-known MIME types assume UTF-8 character encoding wherever a character s
 | `video/H265` | `36`
 | `video/VP8` | `37`
 | |
+| `message/x.rsocket.tracing-zipkin.v0` | `125`
 | `message/x.rsocket.composite-metadata.v0` | `127`
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Artifacts include:
 - Protocol Extensions
   - [Well-known MIME Types](Extensions/Well-Known-MIME-Types.md)
   - [Composite Metdata](Extensions/CompositeMetadata.md)
+  - [Tracing (Zipkin)](Extensions/Tracing-Zipkin.md)
 - [Motivation for creating this protocol](Motivations.md)
 - [Frequently Asked Questions](FAQ.md)
 


Extensions/Tracing-Zipkin.md
```diff
@@ -0,0 +1,55 @@
+# Tracing (Zipkin) Metadata Extension
+
+_This extension specification is currently incubating._
+
+## Introduction
+Observability and tracing are key requirements for robust and reliable applications.  When using distributed applications connected with RSocket, it's important to propagate metadata about the current logical operations throughout the entire system.  One of the most popular systems for doing this kind of tracing is [Zipkin][z].  This extension specification provides an interoperatble structure for Zipkin metadata payloads to contain tracing information.  It is designed such that systems can efficently communicate span and trace information to a Zipkin server and propagate that information throughout a distributed system.
```
yschimke:
```suggestion
Observability and tracing are key requirements for robust and reliable applications.  When using distributed applications connected with RSocket, it's important to propagate metadata about the current logical operations throughout the entire system.  One of the most popular systems for doing this kind of tracing is [Zipkin][z].  This extension specification provides an interoperable structure for Zipkin metadata payloads to contain tracing information.  It is designed such that systems can efficently communicate span and trace information to a Zipkin server and propagate that information throughout a distributed system.
```

yschimke:
Is there value in a spec like 
https://github.com/w3c/distributed-tracing/blob/master/trace_context/HTTP_HEADER_FORMAT.md

Instead of designing another representation?

Extensions/Tracing-Zipkin.md
```diff
@@ -0,0 +1,55 @@
...
+[rf]: ../Protocol.md#frame-fnf
+[rr]: ../Protocol.md#frame-request-response
+[rs]: ../Protocol.md#frame-request-stream
+
+### Metadata Contents
+```
+     0                   1                   2                   3
+     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+    |            Version            |P|S|D|     Flags   |
```
robertroeser: I think after reading yuri's previous comments the version isn't needed because of the mime-type

Extensions/Tracing-Zipkin.md
```diff
@@ -0,0 +1,55 @@
...
+    +---------------------------------------------------------------+
+    |                                                               |
+    +                        Parent Span ID                         +
+    |                                                               |
+    +-+-+-----------------------------------------------------------+
+    |S|D|
+    +-+-+-----------------------------------------------------------+
+```
+
+* **Version**: (16 bits = max value 65,535) Unsigned 16-bit integer of version number of the tracing payload.
```
robertroeser: I think after reading yuri's previous comments the version number isn't needed and we could use the mime-type for that.

Extensions/Tracing-Zipkin.md
nebhale: I'll update this once I get confirmation on [this](
https://github.com/rsocket/rsocket/pull/255#discussion_r227130485).

nebhale:
@yschimke Picking up from [your comment](
https://github.com/rsocket/rsocket/pull/256#issuecomment-431549909), I had specified it binary blobs because that's what the base Zipkin specification uses.  To map into HTTP, they declare certain headers and Base64 encode it, but that will be much larger than the packed binary representation.

Extensions/Tracing-Zipkin.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+* **Flags**: (3 bits)
+  * (**P**)arent Span Id: Tracing payload contains a parent span id.
+  * (**S**)ampling Decision: Tracing payload contains a sample value. (Exclusive with D flag.)
+  * (**D**)ebug: Tracing payload contains a debug value. (Exclusive with S flag.)
+* **Trace ID**: (128 bits) Unsigned 128-bit integer ID of the trace. Every span in a trace shares this ID.
+* **Span ID**: (64 bits) Unsigned 64-bit integer ID for a particular span. This may or may not be the same as the trace id.
+* **Parent Span ID**: (64 bits) Unsigned 64-bit integer ID for a particular parent span.  This is an optional ID that will only be present on child spans. That is the span without a parent id is considered the root of the trace. (Not present if P flag is not set)
+* **Sampling Decision**: The span should be reported to the tracing system. (Not present if S flag is not set)
```
yschimke: I wonder whether these flags should be combined with the first set into a single byte, and it made clear that the first byte is just the flags.

yschimke:
Hey @adriancole, any feedback on this?  We want a tight encoding for zipkin.

adriancole:
thanks for the ping.. dunno why this was off radar. I'm at a workshop today and flying tomorrow. will try to give some thought then.

yschimke:
LGTM - I think it's worth packing the bits into a single byte at the start, then I'm happy with it.

adriancole:
First thought is that if it is important to save bytes from our single format, it might still be worthwhile using the same order. Even better if we can use exactly the same format as other parsers can be used: https://github.com/openzipkin/b3-propagation#single-header

adriancole:
here is some background reading if folks are interested https://cwiki.apache.org/confluence/display/ZIPKIN/b3+single+header+format

nebhale:
@adriancole The single header is still larger than packed binary simply because Base64 is less dense than binary because it only represents 7 bits per character (as I understand it).  So I’m still confident in my choice just to have the binary as it exists today.

Extensions/Tracing-Zipkin.md
nebhale: @yschimke Because the sampling decision and debug flags are ternary (yes, no, and if unset defer to previous) they can’t be represented by a single bit.  Do you think it’s clear enough what’s going on if I end up with two bits for each of those in the flags section?  That’s quite inconsistent with everything else in the protocol.

adriancole:
hi ben. hex is indeed less. however there is also signal loss. I would
implore you to consider the same data and ideally in the same order even if
you encode without hex. trying to avoid unnecessary cognitive load and help
you reuse our experience.

most notably you are missing the debug sampled state. a simple
implementation is to use a single byte which is hex for all sampled states
as is the same in the single header format. less concerned about if you
want to use bit flags just suggesting if one byte isnt overall easier way
to deal  with this. if so you can reuse our docs too as all it is then is a
layout issue (not using hex etc)

wdyt?

robertroeser:
@nebhale @adriancole  Setting payload size aside for a second- this will work with a flyweight, and as well as only allocate stack variable vs mostly like stored on the heap depending the language. In the context of Java - RSocket is binary, zero copy and has access to Netty's ByteBufs. The hex strings that are created can't be cached because they will be unique. Ben's proposal would write tracing data off-heap, with little-to-no GC pressure vs hex strings which would create garbage to be collected. 

If we're missing data we should add it, but otherwise I'm happy with this.

nebhale:
Updated now with a byte-aligned flags field containing two-bit representations of sampling and debug.  I'd consider this to be an official candidate for ratification.

yschimke:
LGTM

adriancole:
I knew there would be some distractions about hex despite my attempts to
suggest this part shouldn't be the focus.

using the same or conceptually linkable layout allows one to write tests on
existing data. there are a lot of subtleties. also I don't think other
formats maybe were considered yet. I know I was late to the party but seems
less people interested in input than going with what is here.

for those who are. here is what I would have replied with:

unless there are tests it is hard to get format exactly right. in such case
it is usually best to follow others or describe where and why it is
different.

for example finagle has a bit flag binary layout. also census does and it
is almost the same. folks do propagate sampled only (ex only a suggestion
to keep a trace without ids). also there are times when parent is never
propagated, but we do need means to tell if it exists or not (ex convention
of banning id zero etc)

in my experience each time someone has tried to do something completely
without feedback it lacks some detail and very hard to fix here and there.
you may believe you can eyeball this and cover all cases but you might be
the first in history to do that.

anyway this is your thing but I think you are missing an opportunity and
this missed opportunity may well result in propagation bugs which are
terribly difficult to fix once out.

the way to verify would be to look at recent b3 single tests in brave or
envoy and port them to a binary codec for this. then inventory the gaps.
this involves both code around the codec and also layout related ones.

if you want to have this format be the exact thing as a supported b3 binary
format I am interested but we must consider things primarily from the point
of existing work. personally I dont want to redocument something only
arbitrarily different.

adriancole:
opened https://github.com/openzipkin/b3-propagation/issues/31 with notes folks can feel free to take or leave. offer stands for help if you don't want to carry this burden alone

nebhale:
@adriancole I suspect there might be some confusion about the exact goal of this extension specification.  The key thing is that we're trying to define a frame structure for propagating information that already exists.  Effectively we're defining a more RSocket-idiomatic equivalent of the `X-b3-*` headers, **not** trying to define any additional behavior or implement any libraries.  We expect applications that currently have Zipkin data floating around to keep doing what they're doing without any significant change.  However, once they attempt to transfer that data across the RSocket protocol, where we do not have HTTP-style headers, we need to have an interoperable mechanism for doing that.

The net result is that if your application currently has the five pieces of data (trace id, span id, parent span id, sample decision, and debug) we serialize that with perfect fidelity into a binary frame, send that across the wire and deserialize it with perfect fidelity back out at the other end.  What _an application_ chooses to do with that data (and as you point out, the inherent difficulties in implementing consistently) is not within scope of this extension specification and is left to existing libraries to handle consistently at the application level.

adriancole:
You are definitely missing a couple things which are much easier to fix than this discussion has been. I would recommend a "learning ear" as it is possible that others may know more about Zipkin practice than you do.

* A trace ID is usually 64 bit. You'll need to mention how to encode it. Usually, this is padding zeros on the left-hand side assuming it was your intention to have a fixed 128-bit field. The reason this is important to document is that people have different ideas, and since 64-bit is the dominant length you should decide how to handle that, either with a byte marker to say done, or the convention of 64-bits.
* It is currently the case that libraries propagate `X-B3-Sampled: 0` by itself. They also propagate `X-B3-Sampled: 1` and sometimes `X-B3-Flags: 1`. Some (redundantly) pass `X-B3-Sampled: 1` and `X-B3-Flags: 1`. You need to actively highlight that this is prohibited (flags only) or tell people how to handle the rest? ex flags can be sent alone or conventionally all zeros following?

These are the two biggest things I can eyeball as undefined on your variant of the b3 spec. Usually, designing a spec is the small work vs ecosystem cleanup. That cleanup might actually fall on people who aren't on this thread, which is why I still am trying hard to get you to empathize with b3's actual spec!

nebhale:
Spec-ing a way to pad 64 bit trace ids is definitely something missing and needs to be improved.

On the sampled/flags subject I believe this is covered in [the documentation](https://github.com/rsocket/rsocket/blob/master/Extensions/Tracing-Zipkin.md#metadata-contents)

* No `X-B3-Sampled` or `X-B3-Flags`
  The first bit of both **S** and **D** are 0, and the second bit of each are ignored ("Not present if first bit is not set").
* `X-B3-Sampled: 1/0`
  The first bit of **S** is set and the second bit is either set or not (header 1 or 0 respectively).  This covers both the boolean values and "can be absent" from the Zipkin spec.
* `X-B3-Flags: 1/0`
  The first bit of **D** is set and the second bit is either set or not (header 1 or 0 respectively).  This covers both the boolean values and "can be absent" from the Zipkin spec.
* `X-B3-Sampled: 1` and `X-B3-Flags: 1`
  The documentation explicitly states that the **S** and **D** flags are "Exclusive with D flag" and "Exclusive with S flag" respectively, which is how the RSocket spec "actively highlight[s] that this is prohibited".

I believe that within the idioms of the RSocket specification we've covered the sample/debug specification to match what is defined by [the HTTP Tracing documentation](https://zipkin.io/pages/instrumenting.html).

admbanking1 (commented):
> https://github.com/openzipkin/b3-propagation/issues/31#issue-377008316

PR #258, opened 2018-10-30
Well-known MIME Types

nebhale:
This change adds an extension for well-known MIME types that maps the types to integer values for optimized inclusion in payloads.

> MIME types are an integral part of the RSocket and Extension specifications.  However, due to their definitions as strings and the number of times they need to be sent as part of typical interaction, they can be wasteful in their typical form.  Because of this, it's useful to represent well-known MIME types as integer values during transmission.  This behavior does not remove the need or ability in the specifications to declare MIME types as strings.

### DIFF
--- a/Extensions/Well-Known-MIME-Types.md
+++ b/Extensions/Well-Known-MIME-Types.md
@@ -1,0 +1,48 @@
+# Well-known MIME Types
+
+## Introduction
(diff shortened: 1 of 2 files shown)


nebhale:
Note, this does not include the MIME types of any of the other inflight proposals.  They will be added after this first cut is ratified and each of them is ratified in turn.

robertroeser:
list looks good to me

nebhale:
`application/cbor` added.  If @robertroeser can give me an approved review, I'll go back and add numbers and it'll be ready for merge.

nebhale:
Avro and GraphQL added.  IDs issued.  Ready for merge.

PR #259, opened 2018-11-05
64-bit Trace Id

nebhale:
Previously the extension specification assumed that all trace ids would be 128-bit values.  In the discussion of #256 it was highlighted that not only was this variable between 64-bit and 128-bit, but the primary behavior is 64-bit values.  This change updates the specification to add a new T flag that indicates whether the value is a 64-bit value (unset) or a 128-bit value (set).  This removes the need to pad 64-bit values and ensures the efficiency in the data structure.

### DIFF
--- a/Extensions/Tracing-Zipkin.md
+++ b/Extensions/Tracing-Zipkin.md
@@ -8,7 +8,7 @@ Observability and tracing are key requirements for robust and reliable applicati
 [z]: https://zipkin.io
 
 ## Metadata Payload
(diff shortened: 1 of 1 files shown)


Extensions/Tracing-Zipkin.md
```diff
@@ -42,13 +42,14 @@ This metadata type is intended to be used per stream, and not per connection nor
 ```
 
 * **Flags**: (8 bits)
+  * (**T**)race Id Size: Unset indicates that the Trace Id is 64-bit. Set indicates that the Trade Id is 128-bit.
   * (**P**)arent Span Id: Tracing payload contains a parent span id.
   * (**S**)ampling Decision (2 bits): Tracing payload contains a sample value. (Exclusive with D flag.)
     * First bit indicates if a sampling decision is present
-    * Second bit indicates that the span should be reported to the tracing system. (Not present if first bit is not set)
+    * Second bit indicates that the span should be reported to the tracing system. (Ignroed if first bit is not set)
```
yschimke:
```suggestion
    * Second bit indicates that the span should be reported to the tracing system. (Ignored if first bit is not set)
```

PR #271, opened 2019-06-14
add types for hessian, java-object, cloudevents json and rsocket.data

linux-china:
x-hessian & x-java-object are very important for Java applications.  cloudevents+json is very promising type for event processing and already used in different systems.  x.rsocket.data-encoding.v0  is for data encoding for every Payload.

### DIFF
--- a/Extensions/WellKnownMimeTypes.md
+++ b/Extensions/WellKnownMimeTypes.md
@@ -50,7 +50,6 @@ All well-known MIME types assume UTF-8 character encoding wherever a character s
 | `application/x-java-object` | `0x27`
 | `application/cloudevents+json` | `0x28`
 | |
-| `message/x.rsocket.data-encoding.v0` | `0x7C`
 | `message/x.rsocket.tracing-zipkin.v0` | `0x7D`
 | `message/x.rsocket.routing.v0` | `0x7E`
 | `message/x.rsocket.composite-metadata.v0` | `0x7F`
--- a/Extensions/WellKnownMimeTypes.md
+++ b/Extensions/WellKnownMimeTypes.md
@@ -46,7 +46,11 @@ All well-known MIME types assume UTF-8 character encoding wherever a character s
 | `video/H264` | `0x23`
 | `video/H265` | `0x24`
 | `video/VP8` | `0x25`
+| `application/x-hessian` | `0x26`
+| `application/x-java-object` | `0x27`
+| `application/cloudevents+json` | `0x28`
 | |
+| `message/x.rsocket.data-encoding.v0` | `0x7C`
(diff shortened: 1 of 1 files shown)


robertroeser:
I'm ok with hessian and java serialization in the composite metadata. I think there is much more likely case that composite metadata will have those variables present then say something like ogg or mp3, etc. If we're worried about the space we should probably get rid of some of the multi-media types as I don't forsee them being used as metadata in a request let alot part of composite metadata.

robertroeser:
@linux-china what does  | `message/x.rsocket.data-encoding.v0` | `0x7C`  signify? I got the other ones

linux-china:
> @linux-china what does | `message/x.rsocket.data-encoding.v0` | `0x7C` signify? I get the other ones

it's the data encoding medatadata for per Payload.  Now RSocket uses global data MIME type, but for some cases, it's not enough.  For example,  App created a connection with broker with global data MIME type application/json, but the  RSocket Cache service in the App reuses the connection to broker and talk with Cache service provider with different data encoding type(hessian, java object etc), and the service provider will return the payload with data encoding metadata included, and cache service client will use data encoding metadata to decode the data. 

One global rsocket connection with different data MIME type support. if no data encoding metadata inlcuded in Payload, use global data MIME type. if present, use data encoding metadata in Payload.

linux-china:
> @linux-china what does | `message/x.rsocket.data-encoding.v0` | `0x7C` signify? I get the other ones

Please correct it if naming is not good :)

robertroeser:
isn't that covered by `message/x.rsocket.composite-metadata.v0` | `0x7F` or is this something different?

linux-china:
> isn't that covered by `message/x.rsocket.composite-metadata.v0` | `0x7F` or is this something different?

Composite metadata is like group for metadata, and message/x.rsocket.data-encoding.v0 is like 
message/x.rsocket.routing.v0 or message/x.rsocket.tracing-zipkin.v0 , and it's one of well known types in composite metadata.   For example, a cache request/response, a  request payload with composite metadata format to include routing and data encoding metadata.   Maybe I misunderstand?

robertroeser:
You get one metadata per connection - so if you want 1 or more different types than - that is what the composite metadata is for, and the well known mime-types.

robertroeser:
@linux-china If you remove `message/x.rsocket.data-encoding.v0 | 0x7C` I will merge the pull request

linux-china:
> @linux-china If you remove `message/x.rsocket.data-encoding.v0 | 0x7C` I will merge the pull request

Removed the data encoding and discuss later

linux-china:
> @linux-china If you remove `message/x.rsocket.data-encoding.v0 | 0x7C` I will merge the pull request

Please add new types on RSocket-Java, thanks.

Issue #272, opened 2019-08-29
Add Security Metadata Extension


rwinch:
I'd like to propose creating standard metadata extensions for the purpose of authenticating users. I'd like to ensure we have support for both username/password and bearer token based credentials.

I think this would include two different metadata MIME Types, but am open to suggestions. Below is a first attempt at defining the support I have in mind.

# Valid Frame Types

Both metadata types can be used for either/or a connection or a stream. This will allow authentication to happen at the time of connection and/or per stream. This is important because authentication/authorization might happen on a per connection basis. For example, a mobile device might only provide credentials at the time of connection.

However, a web application might have authentication/authorization required to establish a connection and then rely on specific user level credentials (i.e. an OAuth Token from the web application user) for authentication/authorization on each stream.

# Basic

The first metadata MIME Type I'd like to propose is `message/x.rsocket.authentication.basic.v0`. The metadata contains a username and password in the format of `<username-length-in-bytes><username-string><password>`.

# Bearer

The second metadata MIME Type I'd like to propose is `message/x.rsocket.authentication.bearer.v0`. The metadata contains a bearer token in the format of `<bearer-token-string>`. In this case a bearer token is defined as a string that when presented grants the "bearer" of the token access so a resource. This conforms with the [OAuth definition of bearer tokens](https://tools.ietf.org/html/rfc6750#section-1.2), but is also applicable for in a broader sense (i.e. session ids).

yschimke:
I’m slightly biased against binary encoding in the first case. But I suspect in JS environments using the bearer token is probably more common. Could the HTTP basic encoding also work for the first case?

rwinch:
Thanks for the prompt response.

> I’m slightly biased against binary encoding in the first case.

I'm far from a JavaScript expert, but can't it also support binary encoding? To be clear I'm open to changes in the proposal, but I'm not certain I understand the hangup of using binary encoding.

> But I suspect in JS environments using the bearer token is probably more common.

I think there are cases where JavaScript might want to use username/password based credentials. For example, the client might use a username/password grant to establish an OAuth token in the first place.

> Could the HTTP basic encoding also work for the first case?

I initially considered this, but it has a few drawbacks.

* HTTP Basic does not allow a `:` in the username field. While this is probably not an issue in practice, it seems to be an unnecessary limitation.
* HTTP Basic will Base64 encode the credentials to encode non-HTTP-compatible characters. This encoding is unnecessary in RSocket.
[...]

PR #275, opened 2019-09-23
Add Authentication Metadata Extension

rwinch:
Fixes gh-272

### DIFF
--- a/Extensions/Security/Authentication.md
+++ b/Extensions/Security/Authentication.md
@@ -26,8 +26,8 @@ This metadata type can be used in a per connection or per stream, and not indivi
     +---------------+-----------------------------------------------+
 ```
 
-* (**A**)uthentication Type: Authentication type is a well known value represented by a unique integer.
-* **Auth ID/Length**: (7 bits = max value 2^7 = 128) Unsigned 7-bit integer.  If A flag is set, indicates a [Well-known Auth Type ID][wk].  If A flag is not set, indicates the Authentication Type Length in bytes.
+* (**A**)uthentication Type: Authentication type is a well known value represented by a unique integer.  If A flag is set (a value of `1`), indicates a [Well-known Auth Type ID][wk].  If A flag is not set (a value of `0`), indicates the Authentication Type Length in bytes.
+* **Auth ID/Length**: (7 bits = max value 2^7 = 128) Unsigned 7-bit integer.  If A flag is set (a value of `1`), indicates a [Well-known Auth Type ID][wk].  If A flag is not set (a value of `0`), indicates the Authentication Type Length in bytes.
 * **Authentication Type**: the type of authentication encoding. This SHOULD be a US-ASCII string.  The string MUST NOT be null terminated.  (Not present if A flag is set)
 * **Authentication Payload**: The authentication payload encoded as defined by the Authentication Encoding Type.
 
--- a/Extensions/Security/Authentication.md
+++ b/Extensions/Security/Authentication.md
@@ -6,10 +6,9 @@ _This extension specification is currently incubating.  While incubating the ver
 Authentication is a necessary component to any real world application. This extension specification provides a standardized mechanism for including both the type of credentials and the credentials in metadata payloads.
 
 ## Metadata Payload
-This metadata type can be used in a per connection or per stream, and not individual payloads and as such it **MUST** only be used in frame types used to initiate interactions and payloads.  This includes [`SETUP`][s], [`REQUEST_FNF`][rf], [`REQUEST_RESPONSE`][rr], [`REQUEST_STREAM`][rs], [`REQUEST_CHANNEL`][rc], and [`PAYLOAD`][p].  The Metadata MIME Type is `message/x.rsocket.authentication.v0`.
+This metadata type can be used in a per connection or per stream, and not individual payloads and as such it **MUST** only be used in frame types used to initiate interactions and payloads.  This includes [`SETUP`][s], [`REQUEST_FNF`][rf], [`REQUEST_RESPONSE`][rr], [`REQUEST_STREAM`][rs], and [`REQUEST_CHANNEL`][rc].  The Metadata MIME Type is `message/x.rsocket.authentication.v0`.
 
 [s]:  ../../Protocol.md#frame-setup
-[p]:  ../../Protocol.md#frame-payload
 [rc]: ../../Protocol.md#frame-request-channel
 [rf]: ../../Protocol.md#frame-fnf
 [rr]: ../../Protocol.md#frame-request-response
--- a/Extensions/Security/Authentication.md
+++ b/Extensions/Security/Authentication.md
@@ -1,0 +1,35 @@
+# Authentication Extension
+
+_This extension specification is currently incubating.  While incubating the version is 0._
+
+## Introduction
+Authentication is a necessary component to any real world application. This extension specification provides a standardized mechanism for including both the type of credentials and the credentials in metadata payloads.
+
+## Metadata Payload
+This metadata type can be used in a per connection or per stream, and not individual payloads and as such it **MUST** only be used in frame types used to initiate interactions and payloads.  This includes [`SETUP`][s], [`REQUEST_FNF`][rf], [`REQUEST_RESPONSE`][rr], [`REQUEST_STREAM`][rs], [`REQUEST_CHANNEL`][rc], and [`PAYLOAD`][p].  The Metadata MIME Type is `message/x.rsocket.authentication.v0`.
+
+[s]:  ../../Protocol.md#frame-setup
+[p]:  ../../Protocol.md#frame-payload
+[rc]: ../../Protocol.md#frame-request-channel
+[rf]: ../../Protocol.md#frame-fnf
+[rr]: ../../Protocol.md#frame-request-response
+[rs]: ../../Protocol.md#frame-request-stream
+
+
+### Metadata Contents
+```
+     0                   1                   2                   3
+     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+    |A| Auth ID/Len |   Authentication Type                        ...
+    +---------------+---------------+---------------+---------------+
+    |                     Authentication Payload                   ...
+    +---------------+-----------------------------------------------+
+```
+
+* (**A**)uthentication Type: Authentication type is a well known value represented by a unique integer.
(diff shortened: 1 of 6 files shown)


Extensions/Security/Basic.md
```diff
@@ -0,0 +1,29 @@
...
+     0                   1                   2                   3
+     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+    |Username Length|     Username  |    Password              ...
+    +---------------+-----------------------------------------------+
+```
+
+* **Username Length**: Username Length in bytes.
+* **Username**:  The UTF-8 encoded username.  The string MUST NOT be null terminated.
+* **Password**:  The UTF-8 encoded password.  The string MUST NOT be null terminated.
```
spencergibb: Password Length?

linux-china: How about tagging for username and password?  https://github.com/rsocket/rsocket/blob/master/Extensions/Routing.md .  The first tag is user name, and the second is password?

rwinch: > Password Length?

Isn't that just the remaining bytes?

OlegDokuka:
I'm thinking will not be it simpler to having something similar to what there is in HTTP, e.g having a type of auth as a part of metadata, which limits the number of possible metadata types to a minimum and does not require further extending to the extension. 

for example: 

```
    |Auth ID/Length |         Auth Type                            ...
    +---------------+-----------------------------------------------+
	|   					Encoded Credentials					   ...
```

where Auth ID could be additionally listed (as for WellKnown MimeTypes) or it could be any Auth Type encoded as a string

OlegDokuka:
Hey, @rwinch @spencergibb 

Do you have any updates?

Regards,
Oleh

rwinch:
@OlegDokuka and I spoke about the changes offline. The goal is to avoid having a metadata extension for each type of authentication. However, I believe we should avoid using constructs that work around limitations in HTTP (i.e. `base64(username:password)` in RSocket.

We have defined something similar to Composite Metadata Extension. However, it is specific to Authentication. This allows for custom encoding of the credentials without the need for a custom mime type.

I have pushed the changes for review.

OlegDokuka:
@linux-china @spencergibb @rdegnan @robertroeser @nebhale @stevegury, @yschimke  folks, do you have any comments/objections?

yschimke:
No strong objections. Happy for those using it to prove it in v0 and burn a version if there are mistakes. 

General concerns: doesn’t cover challenges from HTTP and not sure the optimisation is worth having to invent our own non standard encodings.

Extensions/Security/Authentication.md
```diff
@@ -0,0 +1,35 @@
...
+     0                   1                   2                   3
+     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+    |A| Auth ID/Len |   Authentication Type                        ...
+    +---------------+---------------+---------------+---------------+
+    |                     Authentication Payload                   ...
+    +---------------+-----------------------------------------------+
+```
+
+* (**A**)uthentication Type: Authentication type is a well known value represented by a unique integer.
```
stevegury: It's not clear which bit value does what.
Maybe we should be more explicit, i.e. `0 means Auth Length, 1 means Auth Id`.

rwinch:
@yschimke 

> doesn’t cover challenges from HTTP

I'm sorry. I don't understand. Can you please expand on what you mean by this?

> not sure the optimisation is worth having to invent our own non standard encodings.

How would you encode it then? 

The way a username/password is encoded when using HTTP are based on context. HTTP Basic is base64 encoded because it makes sense for HTTP headers. When a username/password is passed in the body of the request as form parameters (I'd argue this is the most common way username/password is submitted with HTTP), they are URL encoded because that is what makes sense for a form submission. If credentials are submitted as JSON, then they are encoded yet another way.

In short, the way a username/password is encoded when using HTTP is based upon the context. In the same manner, when using RSocket a username/password should be encoded in a way that makes sense for RSocket.

yschimke:
> I'm sorry. I don't understand. Can you please expand on what you mean by this?

This spec is obviously inspired by existing HTTP standards but doesn't consider the full flow of challenge response. 

Instead its focus on prior knowledge usage. 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication

yschimke:
> In short, the way a username/password is encoded when using HTTP is based upon the context.

That's not true at all. There is a spec for basic auth including an encoding. 

https://en.m.wikipedia.org/wiki/Basic_access_authentication

rwinch:
> This spec is obviously inspired by existing HTTP standards but doesn't consider the full flow of challenge response.

Are you pointing out that RSocket does not define how to request a user provide credentials? If so, I agree.

@OlegDokuka and I discussed this offline and I believe he was planning on adding a separate ticket to discuss how to handle errors for authentication required, invalid authentication, and not authorized.

I personally felt that including credentials and credential error handling would be separate specifications that eventually link to one another. This would allow for reusing the error handling when including the credentials in a different way. If we want to combine this into a single specification, I'd be open to discussing that. Just let me know.

> That's not true at all for basic auth. There is a spec for basic auth including an encoding.

HTTP Basic Auth defines an encoding, but it also defines the context - an HTTP header field. Username/password included in a form post, JSON, etc do not use the same encoding as HTTP Basic Auth because they are not included in the header. Instead, they use the appropriate encoding for their contexts. Just as a form submission does not use HTTP basic encoding, RSocket should use an encoding that makes sense for the context of metadata.

> Maybe just rename it to username and password instead of basic.

I agree the name basic is not a good choice due to the fact that many will confuse it with HTTP Basic. My hope is that this is differentiated by HTTP Basic Auth vs RSocket basic, but this is still not as clear as I'd prefer.

I'm personally not a fan of username and password as it seems pretty verbose. Perhaps I am alone with this though. If anyone has thoughts on naming, I'd like to hear it.

Extensions/Security/Authentication.md
```diff
@@ -0,0 +1,35 @@
+# Authentication Extension
+
+_This extension specification is currently incubating.  While incubating the version is 0._
+
+## Introduction
+Authentication is a necessary component to any real world application. This extension specification provides a standardized mechanism for including both the type of credentials and the credentials in metadata payloads.
+
+## Metadata Payload
+This metadata type can be used in a per connection or per stream, and not individual payloads and as such it **MUST** only be used in frame types used to initiate interactions and payloads.  This includes [`SETUP`][s], [`REQUEST_FNF`][rf], [`REQUEST_RESPONSE`][rr], [`REQUEST_STREAM`][rs], [`REQUEST_CHANNEL`][rc], and [`PAYLOAD`][p].  The Metadata MIME Type is `message/x.rsocket.authentication.v0`.
```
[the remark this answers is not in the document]
OlegDokuka: Looks like it is a common part of every extension definition -> for instance https://github.com/rsocket/rsocket/blob/master/Extensions/Routing.md#metadata-payload

OlegDokuka:
> @OlegDokuka and I discussed this offline and I believe he was planning on adding a separate ticket to discuss how to handle errors for authentication required, invalid authentication, and not authorized.

yes, I’m planning to rise another ticket for that

> I'm personally not a fan of username and password as it seems pretty verbose. Perhaps I am alone with this though. Perhaps we could use simple? If anyone has thoughts on naming, I'd like to hear it.

Yeah, I like `simple` for that purpose

Extensions/Security/Authentication.md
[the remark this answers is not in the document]
rwinch: @yschimke You are right that we should not list `PAYLOAD` here because it is not a start of a stream or connection. I pushed https://github.com/rsocket/rsocket/pull/275/commits/fa1a80b9212816b2efce0c538d06ba5366d1a32b to address this.

I kept the commit separate, so it was easier to see the changes. I can clean up the commits later if desired.

Extensions/Security/Authentication.md
rwinch: @stevegury Thanks for the feedback. I tried to follow the conventions I found in the existing specs. The clarification is on the next line just as it is done in [Composite Metadata](https://github.com/rsocket/rsocket/blob/7278d1dd7963893be77069bb3b699489d65f42ac/Extensions/CompositeMetadata.md#metadata-contents):

> If A flag is set, indicates a [Well-known Auth Type ID][wk].  If A flag is not set, indicates the Authentication Type Length in bytes.

I pushed a fix in https://github.com/rsocket/rsocket/pull/275/commits/de71eaac27ee5b9d231515a9662eed020c8b09a3 I kept the commit separate, so it was easier to see the changes. I can clean up the commits later if desired.

If we decide to stick with these changes, I'd suggest we also update Composite Metadata.

rwinch:
> Yeah, I like simple for that purpose

I pushed a change for this https://github.com/rsocket/rsocket/pull/275/commits/732ad717e6f818ab7bed00594038cf16c630bc1a I kept the commit separate, so it was easier to see the changes. I can clean up the commits later if desired.

yschimke:
To be clear. No intention from me to block forward progress. Just feedback to consider. Happy to ship this. Usage trumps drive by opinions :)

Extensions/Security/Authentication.md
```diff
@@ -0,0 +1,34 @@
...
+     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+    |A| Auth ID/Len |   Authentication Type                        ...
+    +---------------+---------------+---------------+---------------+
+    |                     Authentication Payload                   ...
+    +---------------+-----------------------------------------------+
+```
+
+* (**A**)uthentication Type: Authentication type is a well known value represented by a unique integer.  If A flag is set (a value of `1`), indicates a [Well-known Auth Type ID][wk].  If A flag is not set (a value of `0`), indicates the Authentication Type Length in bytes.
+* **Auth ID/Length**: (7 bits = max value 2^7 = 128) Unsigned 7-bit integer.  If A flag is set (a value of `1`), indicates a [Well-known Auth Type ID][wk].  If A flag is not set (a value of `0`), indicates the Authentication Type Length in bytes.
```
OlegDokuka:
```suggestion
* **Auth ID/Length**: (7 bits = max value 2^7 = 128) Unsigned 7-bit integer.  Applicable only if A flag is not set (a value of `0`).
```

Issue #279, opened 2019-11-19
Stream ID 0 frames prioritization


OlegDokuka:
# Motivation 

In many cases, frames prioritization is a challenging task, especially when it comes to the specific application's level requests. However, there is a list of highly important internal, RSocket specific frames, which should be considered as highly prioritized payloads/frames within the RSocket ecosystem. For example, most of the Stream ID 0 frames like KEEPALIVE frame or LEASE brings the most important information about the state of connection/capacity. Therefore, it would make sense to send them with the highest priority or frankly speaking, as soon as possible on the level of certain implementation of RSocket spec.

# Proposal 

I propose to explicitly suggest/recommend in the spec to send Stream ID 0 with the highest priority to ensure that they are delivered with the minimum delay.

Just recently I have created a [PR to RSocket-Java](https://github.com/rsocket/rsocket-java/pull/718) which shows performance impact-less solution for that problem.

yschimke:
My thoughts

1) The spec needs to address stream priorities generally, I guess this is currently "QoS and Prioritization".
2) I think implicit ordering should have previously meant that all frames were sent in generally the order they were received at the duplex connection level.
3) If we prioritise Stream ID 0 traffic, effectively LEASE, KEEPALIVE, METADATA_PUSH, we break the ordering above per stream...
4) ... Should we make it clear that per stream is freely prioritised by the client and sender.  e.g. rsocket-java could support an API to bump the priority of the stream outside the spec.  This maybe what the section below is meant to convey.

But overall, I think this change functionally makes sense, since you definitely want to prioritise KEEPALIVES over a large video stream of effectively infinite PAYLOAD frames etc.


> 
> QoS and Prioritization
> 
> Quality of Service and Prioritization of streams are considered application or network layer concerns and are better dealt with at those layers. The metadata capabilities, including METADATA_PUSH, are tools that applications can use for effective prioritization. DiffServ via IP QoS are best handled by the underlying network layer protocols.

OlegDokuka:
1) Agree
2) True
3) Not sure. Since the order of All the packets within the logical stream are unchanged and Stream ID 0 packets have no direct relation to a logical stream
4) 
5) I think soo

OlegDokuka:
#280 solved by

PR #280, opened 2019-11-22
Provides addition to QoS and Prioritisation

OlegDokuka:
This PR provides an update to the spec, which mentions a Recommendation for all implementations. Basically, it adds a note on prioritized delivering of all the *Stream ID 0* frames in case it is possible.

For example, in RSocket-Java, we have a [Queue](https://github.com/rsocket/rsocket-java/blob/develop/rsocket-core/src/main/java/io/rsocket/internal/UnboundedProcessor.java), which basically enqueues all the frames on the Requester level first, and the underlying duplex connection drains them and delivers later on. It means that the KeepAlive frame as an important one could be [stacked](https://github.com/rsocket/rsocket-java/pull/718) in that `Queue` and delayed in its delivery to the connection and subsequently to the Responder.

Undelivered on time KeepAlive frame could cause unwanted connection close event, which in that case is a false-positive action.

The same relates to Lease, Metadata_Push frames which bring emerging information about peers connected via the network.

See also #279

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -1228,10 +1228,9 @@ A Responder that receives a REQUEST that it can not honor due to LEASE restricti
 <a name="flow-control-qos"></a>
 #### QoS and Prioritization
 
-Quality of Service and Prioritization of all *non Stream ID 0* streams (all application level streams) are considered application or network layer concerns and are better dealt with
-at those layers. The metadata capabilities, including METADATA_PUSH, are tools that applications can use for effective prioritization.
+Quality of Service and Prioritization of streams are considered application or network layer concerns and are better dealt with at those layers. The metadata capabilities, including METADATA_PUSH, are tools that applications can use for effective prioritization.
 
-However, all the frames within the *Stream ID 0* **SHOULD** be considered as *highly-prioritized* frames and delivered to the connection as soon as possible, since all the related frames impact directly or indirectly the stability and performance of the system in total.
+Within a single stream, the frames have to be processed in the order. However, frames with *Stream ID 0* **SHOULD** have a higher priority depending on the implementation, since all those frames impact directly or indirectly the stability and performance of the system in total.
 
 DiffServ via IP QoS are best handled by the underlying network layer protocols.
 
--- a/Protocol.md
+++ b/Protocol.md
@@ -1228,8 +1228,11 @@ A Responder that receives a REQUEST that it can not honor due to LEASE restricti
 <a name="flow-control-qos"></a>
 #### QoS and Prioritization
 
-Quality of Service and Prioritization of streams are considered application or network layer concerns and are better dealt with
+Quality of Service and Prioritization of all non zero streams (all application level streams) are considered application or network layer concerns and are better dealt with
 at those layers. The metadata capabilities, including METADATA_PUSH, are tools that applications can use for effective prioritization.
+
+However, all the frames within the *Stream ID 0* **SHOULD** be considered as *highly-prioritized* frames and delivered to the connection as soon as possible. That is not necessarily a mandated part of the implementation, but a RECOMENDED one since all the related frames impact the stability and performance of the system in total.
+
 DiffServ via IP QoS are best handled by the underlying network layer protocols.
 
 ### Handling the Unexpected
--- a/Protocol.md
+++ b/Protocol.md
@@ -1230,7 +1230,7 @@ A Responder that receives a REQUEST that it can not honor due to LEASE restricti
 
 Quality of Service and Prioritization of streams are considered application or network layer concerns and are better dealt with at those layers. The metadata capabilities, including METADATA_PUSH, are tools that applications can use for effective prioritization.
 
-Within a single stream, the frames have to be processed in the order. However, frames with *Stream ID 0* **SHOULD** have a higher priority depending on the implementation, since all those frames impact directly or indirectly the stability and performance of the system in total.
+Within a single stream, the frames have to be processed in order, but this is not the case between different streams. Frames with *Stream ID 0* **SHOULD** have a higher priority (*this is implementation specific*), since all those frames may impact the performance of the system.
 
 DiffServ via IP QoS are best handled by the underlying network layer protocols.
 
--- a/Protocol.md
+++ b/Protocol.md
@@ -1228,7 +1228,7 @@ A Responder that receives a REQUEST that it can not honor due to LEASE restricti
 <a name="flow-control-qos"></a>
 #### QoS and Prioritization
 
-Quality of Service and Prioritization of all non zero streams (all application level streams) are considered application or network layer concerns and are better dealt with
+Quality of Service and Prioritization of all *non Stream ID 0* streams (all application level streams) are considered application or network layer concerns and are better dealt with
 at those layers. The metadata capabilities, including METADATA_PUSH, are tools that applications can use for effective prioritization.
 
 However, all the frames within the *Stream ID 0* **SHOULD** be considered as *highly-prioritized* frames and delivered to the connection as soon as possible. That is not necessarily a mandated part of the implementation, but a RECOMENDED one since all the related frames impact the stability and performance of the system in total.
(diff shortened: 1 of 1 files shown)


Protocol.md
```diff
@@ -1228,8 +1228,11 @@ A Responder that receives a REQUEST that it can not honor due to LEASE restricti
 <a name="flow-control-qos"></a>
 #### QoS and Prioritization
 
-Quality of Service and Prioritization of streams are considered application or network layer concerns and are better dealt with
+Quality of Service and Prioritization of all *non Stream ID 0* streams (all application level streams) are considered application or network layer concerns and are better dealt with
 at those layers. The metadata capabilities, including METADATA_PUSH, are tools that applications can use for effective prioritization.
+
+However, all the frames within the *Stream ID 0* **SHOULD** be considered as *highly-prioritized* frames and delivered to the connection as soon as possible. That is not necessarily a mandated part of the implementation, but a RECOMENDED one since all the related frames impact the stability and performance of the system in total.
```
yschimke: > That is not necessarily a mandated part of the implementation, but a RECOMENDED one since all the related frames impact the stability and performance of the system in total.

I think SHOULD covers this.

linux-china: yes. Metadata_Push is very important. Now we use Metadata_Push to broadcast network changes of rsocket broker cluster.

OlegDokuka:
@rdegnan @nebhale @smaldini @stevegury, folks, please let me know that you are aware of that enhancement and good/not-good with the proposed change.

Regards,
Oleh

Protocol.md
yschimke: @OlegDokuka to clarify my point, I'd delete that sentence and just rely on SHOULD or MUST.

OlegDokuka: @yschimke, fixed! Thanks for suggestion ❤️ 🙏

OlegDokuka:
@nebhale, @smaldini, @stevegury, @rdegnan, folks, please give a little of your time to this PR :D

stevegury:
I think `MUST` is too strong here, this should be a `SHOULD`. I believe this should be an implementation recommendation at best.

OlegDokuka:
> I think `MUST` is too strong here, this should be a `SHOULD`. I believe this should be an implementation recommendation at best.

Yeah, sounds good to me

stevegury:
I am a little bit concerned that we're opening the pandora box here. We are using the word `highly-prioritized` without clearly defining it. But even if we'd have defined it, we don't want to be too prescriptive here.

Maybe we could keep the spec simple by just stating that stream priorities can exist, but we do not prescribe any implementation (I think this was, in essence, what the previous section about QoS was saying).

Maybe something like this:
> Quality of Service and Prioritization of streams are considered application or network layer concerns and are better dealt with at those layers. The metadata capabilities, including METADATA_PUSH, are tools that applications can use for effective prioritization.
> 
> Within a single stream, the frames have to be processed in order. However, between different streams, frames can be processed out of the order they have been received. (For instance, frames from the stream 0 can be processed with a higher priority)

OlegDokuka:
right now in Java we have a list of issues related to missvaluing frames under stream ID 0 like keepalive which leads to unwonted crashes. In turn, spec says nothing about streams produced by user which means others streams are unchenged and how to prioritize them is on user.

I'm afraid, if we are not going to specify somewhere that stream ID 0 streams *should*  be sent as soon as possible we will have unstebale implementation of rsocket itself

yschimke:
Agreed. A well behaved client that prioritises keepalives on stream 0 is still susceptible to issues with a server that doesn’t.

stevegury:
I understand your valid concerns.

But I believe the problem is that we may be confounding application level keep-alive with transport level keep-alive. If your server is not able to processed a keep-alive, and/or your not able to processed the response within the allocated time, it probably means that either the allocated time is too small or that the server is too overloaded to provide satisfactory prompt responses (then it's ok to close the connection, that's the purpose of keep-alive).

The specification mentioned that `a reasonable time interval between client KEEPALIVE frames is 500ms.`, I always considered that value to be quite aggressive. If you use it, it means that you're ok considering the connection as dead if the server (+network) latency goes above that value (even for a short period).

That being said, the current spec as is, or with the modification I'm proposing, doesn't prevent you to implement a higher priority for the stream 0.

yschimke:
I’m not sure it’s true that this is strictly about overloaded servers or connections.  It merely requires enough content on a non 0 stream to consume the available bandwidth (which may be limited eg mobile) for a small multiple of the keepalive time. So a 100mb file transfer on 3G.

OlegDokuka:
@stevegury, let me give you a couple of samples where behaviors are expected and not:

## expected:

1) Normal data sending


The image above shows the most common case, we send frames, and we receive them. The frame rate is more than low to let the network deliver them, and the server processes them on time. 

2) Normal data sending with outage


The image above shows the case where we got network outage, so most parts of the frame started dropping, so after some time of not delivering any frames at all, receiver fires keepalive error to close the connection immediately and notify the producer in that way that the connection is in the bad state. 

## unexpected:


In the image above, we have a connection in a good state. However, the Quality of serving the response to incoming frames became lower since the rate on the requester side is high enough to overwhelm the responder. In such a case in the current setup in RSocket-Java, we will receive **keepalive error**. That is something unexpected because the connection is in the good state, and the server can process the responses, but not in the rate of the producer. It means that the expected behavior is the producer receive timeout exception on a set of requests (or frames), but not the connection closed exception.

## How should it works instead (in my opinion)

I still believe, in case the user decided to overwhelm the recipient, then the worth that the recipient should expect is timeout exceptions. If the recipient is alive, it will send back keepalive but will not be able to respond within defined SLA.

if the **recipient** throws the keepalive error - it should mean that quality of the connection is soo low so we cannot keep with that anymore and we have to tune the network somehow

if the **producer** throws the keepalive error, it means that the recipient is in the outage state so it can not respond at all and we have to kill it

That is my thoughts on how the expected behaviors.

stevegury:
@yschimke you have a fair example, and this is why fragmentation was introduced in the specification. The 100MB message should not "head-of-line" block the sending of the keepalive frame on stream 0. On the server side, the processing of the keepalive can be done as soon as you receive the frame, you don't need to wait for the large payload on stream `n` to complete.

To clarify my point, I think it's ok to treat stream 0 with a higher priority, I would just like to avoid the pitfall of defining precisely what "higher-priority" means and let the different implementations choose how it's made. That's the purpose of the second paragraph I proposed before, we just state the requirement of a stream, i.e. frames have to be processed in order within the stream but not between stream.

Protocol.md
```diff
@@ -1228,10 +1228,9 @@ A Responder that receives a REQUEST that it can not honor due to LEASE restricti
...
 #### QoS and Prioritization
 
-Quality of Service and Prioritization of all *non Stream ID 0* streams (all application level streams) are considered application or network layer concerns and are better dealt with
-at those layers. The metadata capabilities, including METADATA_PUSH, are tools that applications can use for effective prioritization.
+Quality of Service and Prioritization of streams are considered application or network layer concerns and are better dealt with at those layers. The metadata capabilities, including METADATA_PUSH, are tools that applications can use for effective prioritization.
 
-However, all the frames within the *Stream ID 0* **SHOULD** be considered as *highly-prioritized* frames and delivered to the connection as soon as possible, since all the related frames impact directly or indirectly the stability and performance of the system in total.
+Within a single stream, the frames have to be processed in the order. However, frames with *Stream ID 0* **SHOULD** have a higher priority depending on the implementation, since all those frames impact directly or indirectly the stability and performance of the system in total.
```
stevegury: Look good, I made a few changes below. What do you think of this:
_Within a single stream, the frames have to be processed in order, but this is not the case between different streams. Frames with *Stream ID 0* **SHOULD** have a higher priority (this is implementation specific), since all those frames may impact the performance of the system._
(I removed the word stability since it may means something different for some people).

OlegDokuka: I like it!

OlegDokuka: fixed! ❤️🙏

Issue #281, opened 2019-12-05
Composite Metadata Extension: data Mime/type definition per-stream


OlegDokuka:
# Problematic 

For now, we have the only way to define data mime/type at the very beginning of the connection setup in the `SetupPayload`. For many cases, it is fine to have a homogeneous mime-types (unified "format"), or frankly speaking a mime/type per connection. However, when the case comes to fronted, gateway, or broker, it is necessarily needed to define what data is packed within a logical stream.

# Proposal 

That discussion has already been raised at #117, and it is clear that we can define what data is in the payload using metadata. However, it could be tricky having an unspecified way to do that and, especially today, when we have a [Composite Metadata](https://github.com/rsocket/rsocket/blob/master/Extensions/CompositeMetadata.md) came in, it becomes pretty straight forward on how to solve that problem.
For instance, we can define a specific [WellKnownType](https://github.com/rsocket/rsocket/blob/master/Extensions/WellKnownMimeTypes.md) for doing an override of data mime-type.

1) `SetupPayload` still has a mandatory field for data mime-type, so at the first connection setup, we define a general mime-type.
2) If for interaction within a connection we define metadata as a `CompositMetadata` then it becomes possible to override a data mime per stream
3) If a requester during the interaction does not specify a mime-type using composite metadata, then the default mime-type is a mime-type for that stream
4) If a requester specifies mime-type using composite metadata, then the default mime-type is overridden by one specified in composite metadata for that stream

rstoyanchev:
We had a similar request in https://github.com/spring-projects/spring-framework/issues/23431.

Overriding the data mime type from the `SETUP` frame per stream in composite medatadata seems a straight-forward enough idea. Doesn't that also naturally lead to the possibility for separate input and output data mime types per stream? 

Maybe there could be two possible per-stream overrides, one for input and one for output. If only one is specified, the other assumes the connection default.

OlegDokuka:
Yeah, sounds reasonable, at least proven, but long-lived HTTP. Thus does not have any objection. Work is in progress. I will ping you @rstoyanchev once have something

linux-china:
@OlegDokuka   almost, a stream contains data encoding & accepted encoding, just like a HTTP Request. We can follow the Composite Metadata design with following structure. 
   
    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
[...]

PR #285, opened 2019-12-17
Ability to override Data MIME Type per stream

OlegDokuka:
This PR provides an initial draft (#281) for Metadata extension specification, which allows overriding Request Payload data MIME Type as well as defining Accepted Response Payload Data MIME Types. 

Both parties, Requester and Responder, are allowed to provide Payload Data MIME-type in order to specify a directed series of payload data MIME-Type. 

Responder can or can not consider the proposed list of accepted Payload Data MIME-Types.

If Payload Data MIME-Type is not defined it means that the default mime-type is the one defined by ConnectionSetupPayload


OlegDokuka:
cc/ @rstoyanchev

OlegDokuka:
ping @stevegury @rdegnan @nebhale @smaldini @rstoyanchev @spencergibb

PR #286, opened 2019-12-18
Clarify username length field

OlegDokuka:
This PR clarifies the username length field for Simple Authentication type payload


OlegDokuka:
cc/ @rwinch

PR #287, opened 2019-12-19
Fixes bit length in Simple Auth payload

OlegDokuka:
Looks like it either type in frame representation or frame description :

1) Frame says 8 bits (from 0 bit to 7 bit) where 
2) The description says 7 bits

From my point of view, there is no sense to limit Username length to 128 bits, so it should be fine to have 8 bits for the Username length field


OlegDokuka:
cc/ @rwinch

PR #291, opened 2020-02-07
Minor re-wording of definition for Tag

rstoyanchev:


OlegDokuka:
The only comment is that I'm not sure on removing "the" since we use it everywhere 🤣

rstoyanchev:
> The only comment is that I'm not sure about removing "the" since we use it everywhere

Where are you seeing this? Looking around, including both protocol and extension, I find that field descriptions typically don't start with "the". The ones related to security are the only exception. So for consistency we should probably remove it everywhere. 

I think use of "the" reads pretty well in some cases where it emphasizes it's the one and only, like "the password", but if used consistently, it doesn't sound right in other places, like "(**M**)etadata: The metadata present". By contrast, not using it reads well consistently, and is less distracting.

OlegDokuka:
Yeah. My bad

PR #294, opened 2020-02-10
Cleanup of Routing-And-Forwarding.md

spencergibb:
Fixes tables of contents with links, frame diagrams, tables and a few readability issues.

Because the original files was CRLF, 3f49cc01ab149506838a523056f5dfce80fb9417 will be easier to view differences.


OlegDokuka:
As far as I remember @rdegnan was mentioned to move "broker" spec to a separate repo

PR #302, opened 2020-04-17
Clarify Metadata Length

OlegDokuka:
I would like to add clarification to the metadata length field since right now it seems that 0 metadata length is a valid one but it makes no sense especially when we have a separate metadata flag field to indicate metadata presence.

I guess this wording was missed by mistake when #229 was merged and I don't see any reasons to have metadata flag present along with ZERO metadata encoded (which means 24 bits are just occupied for nothing).

Also, I would like to pay everyone's attention to `2^24-1` which is less convincing when we know for sure 0 is impossible (zero is disallowed, so we can have a little longer range), BUT this is done strictly for *backward* compatibility and in any case, it is impossible to have `2^24-1` bits encoded since metadata length MUST be strictly less than MAX frame length according to #232

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -216,7 +216,7 @@ On a frame that only has Metadata, the Metadata length field is NOT needed:
 ```
 
 
-* __Metadata Length__: (24 bits = max value 16,777,215) Unsigned 24-bit integer representing the length of Metadata in bytes. Excluding Metadata Length field.
+* __Metadata Length__: (24 bits = max value 2^24-1 = max value 16,777,215) Unsigned 24-bit integer representing the length of Metadata in bytes. Excluding Metadata Length field. Value MUST be > 0.
 
 <a name="stream-identifiers"></a>
 ### Stream Identifiers
--- a/Protocol.md
+++ b/Protocol.md
@@ -216,7 +216,7 @@ On a frame that only has Metadata, the Metadata length field is NOT needed:
 ```
 
 
-* __Metadata Length__: (24 bits = max value 2^24-1 = max value 16,777,215) Unsigned 24-bit integer representing the length of Metadata in bytes. Excluding Metadata Length field. Value MUST be > 0.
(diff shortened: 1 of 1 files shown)


OlegDokuka:
@jjeffcaii Actually, that was a kind of root cause of your issue at rsocket java https://github.com/rsocket/rsocket-java/issues/764

jjeffcaii:
@OlegDokuka Yes, I feel that! 😂

rstoyanchev:
I think it is fine for the spec to clarify that when metadata length is 0, the metadata flag should not be set, and therefore metadata not included at all. 

However I see it as an optimization recommendation more than something in need of being explicitly forbidden. There is already a metadata flag for that and implementations have all that is necessary to do the right thing. 

Either way on the receiving end implementations should be lenient, even when the length is 0, and process the frame correctly.

OlegDokuka:
@rstoyanchev That might be true unless `hasMetadata` in `Payload` returns empty buffer which might be confusing. I guess from the implementation perspective (on Java side at least) there are no restrictions on metadata length, so zero-length metadata is processed correctly. However, should it mean anything to a user having Empty metadata encode?

I guess from the user perspective empty metadata should be equal to zero length metadata and should indicate that there is no metadata encoded.

yschimke:
FWIW We had battle over this between java and C++ platforms.  FWIW I think metadata presence is distinct from the length, and making 0 length metadata == missing seems like a bug. All our platforms have types for null vs "", so you are specifically combining these which is a mistake in my opinion.

OlegDokuka:
Alright, maybe I'm mistaken but what is the value of having empty metadata?

yschimke:
> Alright, maybe I'm mistaken but what is the value of having empty metadata?

Special values values in a protocol like this make me nervous.  It seems better to follow established language conventions like `null != ""`.  Even protobuf which defaults an optional string to "", will let you check whether it was set, therefore allow to distinguish between present and "" or missing and null.

OlegDokuka:
Alright. You convinced me. Close that as wontfix. Thanks, everyone for participating

PR #304, opened 2020-05-08
provides aligning with Zipkin b3-propagation

OlegDokuka:
This PR provides alignment with [b3-propagation](https://github.com/openzipkin/b3-propagation#sampling-state-2) and rearrange flags to be following general 4 state support:

all flags are unset 
```
R is 0 S is 0 D is 0 means Deferred (according to sampling)
R is 1 S is 0 D is 0 means Denied (trace should not be sampled)
R is ? S is 1 D is 0 means Accepted (trace can be reported)
R us ? S is ? D is 1 means Debug (trace must be reported)
```


OlegDokuka:
cc @adriancole

adriancole:
I think you understand the tri-state flag. nice!

It is an interesting thing to say `(**R**)eject` I hadn't thought of this. Let's walk through this..

Given Sampled = S, Debug = D, Rejected = R vs Known = K
and bit layouts K,S,D (current except 3 bits) vs R,S,D (proposed)

State         | KSD | RSD | What's being said
---------- | -----|-----|-------------------
Unknown | 000 | 000 | I'm a proxy choosing a trace/span ID for log correlation
Rejected  | 100 | 100 | I'm not tracing this request, and neither should anyone downstream
Sampled  | 110 | 010 | I'm tracing this request, and so should everyone downstream
Debug     | 111 | 001 | Not only should this request be traced, but collectors shouldn't drop it

So far so good. Now, lets fill all bit combinations to see what happens..

Bits | KSD                        | RSD
---- | -------------------|-----------
000 | Unknown (explicit) | Unknown (explicit) 
001 | Debug (implicit)      | Debug (explicit) 
010 | Sampled (implicit)  | Sampled (explicit) 
011 | Debug (implicit)      | Debug (implicit) 
100 | Rejected (explicit)  | Rejected (explicit)
101 | Debug (implicit)      | Nonsense (both Rejected and Debug)
110 | Sampled (explicit)   | Nonsense (both Rejected and Sampled)
111 | Debug (explicit)       | Nonsense (Rejected, Sampled and Debug)

I think you'll agree that while the layout you propose operates fine in ideal case,
someone can make a small mistake and create nonsense values. This isn't possible
in the former layout, which while can be shortened a bit (literally), is otherwise ok.

Incidentally, it is debatable whether Debug bit should be first or last.. Finagle has it first, but the decision tree ends up the same. Notably, Finagle has the sampling bits first, presumably because B3 format does not require trace IDs to be sent along with a sampling decision. Having sampling state first in binary format gives more flexibility on this point.

https://github.com/twitter/finagle/blob/b8523a4541953d6bbc6a64b9409d75ff162afcd9/finagle-core/src/main/scala/com/twitter/finagle/tracing/Flags.scala#L5-L38

OlegDokuka:
@adriancole in fact I meant the same behavior as you have shown for KSD where whatever is put in R or S flags but if D is present it means debug is enabled and the rest simply does not matter. The same valid for S of S is set then if R 0 or 1 it does not matter, we say we have to sample that trace. 

Do you think it is a matter on how we name those bits?

Extensions/Tracing-Zipkin.md
```diff
@@ -44,12 +44,9 @@ This metadata type is intended to be used per stream, and not per connection nor
...
   * (**T**)race Id Size: Unset indicates that the Trace Id is 64-bit. Set indicates that the Trace Id is 128-bit.
   * (**P**)arent Span Id: Tracing payload contains a parent span id.
-  * (**S**)ampling Decision (2 bits): Tracing payload contains a sample value. (Exclusive with D flag.)
-    * First bit indicates if a sampling decision is present
-    * Second bit indicates that the span should be reported to the tracing system. (Ignored if first bit is not set)
-  * (**D**)ebug (2 bits): Tracing payload contains a debug value. (Exclusive with S flag.)
-    * First bit indicates if a debug decision is present
-    * Second bit indicates that the trace should be reported to the tracing system and also override any collection-tier sampling policy. Debug implies an accept sampling decision. (Ignored if the first bit is not set)
+  * (**R**)eject: Tracing payload should not be sampled. (Ignored when S flag or D flag is set.)
+  * (**S**)ampl: Tracing payload should be accepted for tracing. (Ignored when D flag is set.)
```
yschimke, line 48:
```suggestion
  * (**S**)ample: Tracing payload should be accepted for tracing. (Ignored when D flag is set.)
```

yschimke:
LGTM - I'll defer to @nebhale and @adriancole

adriancole:
If you feel strongly to add a negative bit (Rejected = R) considering what I mentioned, you are free to do that.

The former is more conventional and like other binary encodings that have existed for the last 5 years. Changing isn't wrong, just I don't personally see why you would want to create situations of nonsense states intentionally.

adriancole:
sorry 8 years as I think finagle's encoding which the former is based on is probably 8 years old by now.

adriancole:
ps I had a bug in my table.. 101 is nonsense in both

101 | Nonsense (both Rejected and Debug) | Nonsense (both Rejected and Debug)

I think this came up a long time ago when someone was asking about finagle so we had to say that debug wins. It is indeed one exception, but luckily only one

OlegDokuka:
```
101 | Nonsense (both Rejected and Debug) | Nonsense (both Rejected and Debug)
```
is it done for a purpose? 

I mean in my mind Debug should win them all, regardless what is set in `K` and `S` ??1 if Debug is present we say  `state is debug`. Thus I'm wondering whether it makes sense to have that nonsense state?

adriancole:
I don't think we should document the state of all bitflags, only the intended (explicit) ones. I only showed the entire table to highlight how people get this wrong a lot.

The way I'd handle this is I'd add the explicit flags in the README as a table, then in a separate RATIONALE doc show the entire table as while maintainers should know it is distracting to end users to show things they shouldn't be doing.

adriancole:
luckily I think the revisions discussed (assuming bit 4 is ignored) is just editorial in nature. That doesn't dismiss its value. Hours spent between us sorting this out could easily be multiplied across potential readers in the future without your revision.

Keep up the good work, iotw

OlegDokuka:
@adriancole, added an extra flag for the case when sample payload is set or not (sample-set == tracing payload set == tracing payload is present). The reset of the fields remains unchanged with the same behavior, So one can do whatever with them but first of all, on the decoding side we will be looking at the first bit to figure out whether we have to decode anything or not at all.

Extensions/Tracing-Zipkin.md
```diff
@@ -42,14 +42,12 @@ This metadata type is intended to be used per stream, and not per connection nor
 ```
 
 * **Flags**: (8 bits)
+  * (**B**)ody Set: Tracing payload is set.
```
adriancole, line 45:
```suggestion
  * (**I**)Ds Set: When zero, the metadata only includes sampling information and no IDs
    * For example, a health check might set only the (**N**)ot Sampled flag without generating IDs
```

Extensions/Tracing-Zipkin.md
```diff
@@ -42,14 +42,12 @@ This metadata type is intended to be used per stream, and not per connection nor
 ```
 
 * **Flags**: (8 bits)
+  * (**B**)ody Set: Tracing payload is set. 
+  * (**D**)ebug: Tracing payload should be force traced.
+  * (**S**)ample: Tracing payload should be accepted for tracing. (Ignored when D flag is set.)
+  * (**R**)eject: Tracing payload should not be sampled. (Ignored when S flag or D flag is set.)
```
adriancole, line 48: I see the reject notion is sticking with you vs sampling set. If you do that, I think reject is still a bad word to use in flags used for different things. We never use it in practice, we only used 'Reject' in the README to choose a word besides Sampled, and in that case we pair it with Accept. Reject doesn't pair with Sampled and is hence more jargon.

```suggestion
  * (**N**)ot Sampled: Tracing payload should not be sampled. (Ignored when S flag or D flag is set.)
```

I looked and in practice some say unsampled a little more say not sampled.

I think that this is important enough to pick a good name and not blindly follow the text description in the README as I've mentioned that was not trying to pick flag names and if anything we should pick names that are similar to ones in use to cause least confusion, right?

Once we settle on this point I'd like to ping some other zipkin folks as this could well be inherited into our spec.

OlegDokuka:
@adriancole thanks a lot for your effort! ❤️

PR #317, opened 2021-03-02
Fixed a sentence that was not clear

MrMoon:
> This protocol is designed for use both in datacenter, server-to-server use cases, as well as server-to-device use cases over the internet, such as to mobile devices or browsers.

was not very clear on the `for use both in` 
https://github.com/rsocket/rsocket-website/pull/41

### DIFF
--- a/Motivations.md
+++ b/Motivations.md
@@ -132,7 +132,7 @@ RSocket supports session resumption, allowing a simple handshake to resume a cli
 
 RSocket supports two forms of application-level flow control to help protect both client and server resources from being overwhelmed.
 
-This protocol is designed for use both in datacenter, server-to-server, use cases, and server-to-device use cases over the internet, such as to mobile devices or browsers. 
+This protocol is designed to handle the server-server use case, and the client-server use case. This means that it can be used in data centers, in edge network devices, to call other services, over the internet with mobile devices and browsers, and more. 
 
 ##### "Reactive Streams" `request(n)` Async Pull
 
--- a/Motivations.md
+++ b/Motivations.md
@@ -132,7 +132,7 @@ RSocket supports session resumption, allowing a simple handshake to resume a cli
 
 RSocket supports two forms of application-level flow control to help protect both client and server resources from being overwhelmed.
 
-This protocol is designed for use both in datacenter, server-to-server, use cases, as well as server-to-device use cases over the internet, such as to mobile devices or browsers. 
+This protocol is designed for use both in datacenter, server-to-server, use cases, and server-to-device use cases over the internet, such as to mobile devices or browsers. 
 
 ##### "Reactive Streams" `request(n)` Async Pull
 
--- a/Motivations.md
+++ b/Motivations.md
@@ -132,7 +132,7 @@ RSocket supports session resumption, allowing a simple handshake to resume a cli
 
 RSocket supports two forms of application-level flow control to help protect both client and server resources from being overwhelmed.
 
-This protocol is designed to handle the server-server use case, and the client-server use case. This means that it can be used in data centers, in edge network devices, to call other services, over the internet with mobile devices and browsers, and more. 
+This protocol is designed to handle the server-server use case and the server-to-device use case. This means that it can be used in data centers, in edge network devices, to call other services, over the internet with mobile devices and browsers, and more. 
 
 ##### "Reactive Streams" `request(n)` Async Pull
 


OlegDokuka:
@MrMoon can you please rephrase it a little since what you have proposed breaks the idea behind the sentence? 

What this sentence tried to identify by `both` is that the protocol is for server-to-server (a.k.a in-datacenter use case) as well as for device-to-server (a.k.a edge network use-case). Thus if it is not that clear, can we rephrase it in that way so the semantic will not be broken, since right now both sounds to be applicable for datacenter and server to server

MrMoon:
@OlegDokuka does this make it more clear ? or should I try something else

Motivations.md
```diff
@@ -132,7 +132,7 @@ RSocket supports session resumption, allowing a simple handshake to resume a cli
 
 RSocket supports two forms of application-level flow control to help protect both client and server resources from being overwhelmed.
 
-This protocol is designed for use both in datacenter, server-to-server, use cases, as well as server-to-device use cases over the internet, such as to mobile devices or browsers. 
+This protocol is designed to handle the server-to-server (in-datacenter/cloud) use case, and the device-to-server ([edge network](https://www.omnisci.com/technical-glossary/edge-network)) use case. As well as server-to-device use cases over the internet, such as to mobile devices or browsers.
```
yschimke, line 135: I like the change, but the combination of both device-to-server and server-to-device reference is new and also awkward.  Most likely devices connect to server but allows bi-di comms.

MrMoon: should I use examples instead of the device-to-server and server-to-device, so we just remove the server-to-server and the device-to-server, and put examples of these things instead.
and then clear out that these are called device-to-server and server-to-device ?

yschimke: I think generally we have server-server (in DC, all owned by same dev), server-server (hitting other services you don't own) and client-server (mobile use case).

MrMoon:
I think I wrote a better and cleaner sentence,
BTW, why is this sentence in the Flow Control section, wouldn't it be better if it was in the Interaction Models section or in the Polyglot Support. I think it would fit better in the Interaction Models as a first sentence in that section

MrMoon:
Done.
What about the 

> BTW, why is this sentence in the Flow Control section, wouldn't it be better if it was in the Interaction Models section or in the Polyglot Support. I think it would fit better in the Interaction Models as a first sentence in that section

should I also move it somewhere else, or the place of the sentence is good where it is ?

OlegDokuka:
@MrMoon I think your comment is reasonable. How about placing this sentence right after 

```
One such example is push notifications. Using request/response forces an application to do polling where the client consistently sends requests to check the server for data. One does not need to look far to find examples of applications doing high volumes of requests/second just to poll and be told there is nothing for them. This is wasteful for clients, servers, and networks, costs money, increases infrastructure size, operational complexity and thus availability. Generally it also adds latency to the user experience in receiving a notification, as polling is scaled back to longer intervals in an attempt to reduce costs.

```

OlegDokuka:
@MrMoon wait a sec.

Motivations.md
```diff
@@ -132,7 +132,7 @@ RSocket supports session resumption, allowing a simple handshake to resume a cli
 
 RSocket supports two forms of application-level flow control to help protect both client and server resources from being overwhelmed.
 
-This protocol is designed for use both in datacenter, server-to-server, use cases, as well as server-to-device use cases over the internet, such as to mobile devices or browsers. 
+This protocol is designed to handle the server-server use case, and the client-server use case. This means that it can be used in data centers, in edge network devices, to call other services, over the internet with mobile devices and browsers, and more.
```
OlegDokuka, line 135:
```suggestion
This protocol is designed to handle the server-server use case and the server-to-device use case. This means that it can be used in data centers, in edge network devices, to call other services, over the internet with mobile devices and browsers, and more. 
```

OlegDokuka: let's preserve naming

MrMoon:
Do I need to fix that check ?

OlegDokuka:
@MrMoon no worries, it is fine

rstoyanchev:
@MrMoon after a discussion with @OlegDokuka, we concluded the server-to-server and server-to-device references apply specifically to the flow control section where they were meant to how the types of flow control apply to each scenario. We think that those explanations are a bit too narrowly focused on specific scenarios given the symmetrical nature of RSocket and have improved the whole section all around.

MrMoon:
> @MrMoon after a discussion with @OlegDokuka, we concluded the server-to-server and server-to-device references apply specifically to the flow control section where they were meant to how the types of flow control apply to each scenario. We think that those explanations are a bit too narrowly focused on specific scenarios given the symmetrical nature of RSocket and have improved the whole section all around. 

Ohhhhh
Yeah I see it now, and I have a better understanding on how it fits, Good call 👌!!!

PR #320, opened 2021-03-31
improves Fragmentation and Reassembly examples

OlegDokuka:
# Provide better examples for Fragmentation

### Motivation:

At the moment, fragmentation examples do not explain in details how every frame (including flags) should look like

### Modifications:

This PR adds flags to the example sequences as well as add cases when payload has next and complete flags and adds example for Request Channel which has complete flag

### Result:

Improved examples

Signed-off-by: Oleh Dokuka <odokuka@vmware.com>

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -949,11 +949,14 @@ When a PAYLOAD is fragmented, the Metadata MUST be transmitted completely before
 
 For example, a single PAYLOAD with 20MB of Metadata and 25MB of Data that is fragmented into 3 frames:
 
+##### Payload with `next` flag set
+
 ```
 -- PAYLOAD frame 1
 Frame length = 16MB
 (M)etadata present = 1
 (F)ollows = 1 (fragments coming)
+(N)ext = 1
 Metadata Length = 16MB
 
 16MB of METADATA
@@ -963,6 +966,7 @@ Metadata Length = 16MB
 Frame length = 16MB
 (M)etadata present = 1
 (F)ollows = 1 (fragments coming)
+(N)ext = 1
 Metadata Length = 4MB
 
 4MB of METADATA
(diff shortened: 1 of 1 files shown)


Protocol.md
```diff
@@ -1001,6 +1048,43 @@ Metadata Length = 16MB
...
+If the Requester wants to cancel sending a fragmented sequence, it MAY send a CANCEL frame without finishing delivery of the fragments. 
+
+##### Request Channel with `complete` flag
+
+```
+-- REQUEST_CHANNEL frame 1
+Frame length = 16MB
+(M)etadata present = 1
+(F)ollows = 1 (fragments coming)
+(C)omplete = 0
```
dnadoba, line 1076:
```suggestion
(N)ext = 1
(C)omplete = 0
```

dnadoba (commented):
We should add a note that, in the examples, all other flags are assumed to be `0` i.e. not set. Alternatively we could  write out every flag explicitly but this will be very verbose.

Protocol.md
```diff
@@ -949,11 +949,14 @@ When a PAYLOAD is fragmented, the Metadata MUST be transmitted completely before
 
 For example, a single PAYLOAD with 20MB of Metadata and 25MB of Data that is fragmented into 3 frames:
 
+##### Payload with `next` flag set
```
rstoyanchev, line 952: I don't think this needs to be a sub-section with a heading, especially coming after a `":"`. Rather it can be just the first example you see. So basically I suggest removing the sub-heading completely. Then after that, below for the second example, simply add a sentence before it. I'll add a concrete suggestion there.

Protocol.md
```diff
@@ -972,13 +976,54 @@ Metadata Length = 4MB
...
 (F)ollows = 0
+(N)ext = 1
 
 0MB of METADATA
 13MB of Data
 ```
 
 If the sender (Requester or Responder) wants to cancel sending a fragmented sequence, it MAY send a CANCEL frame without finishing delivery of the fragments. 
 
+##### Payload with `next` and `complete` flags set
```
rstoyanchev, line 987:
```suggestion
While in a Request-Stream, `PAYLOAD` frames have only the `(N)ext` flag, in a `Request-Response` the `PAYLOAD` frame will also have a `complete` flag as shown below:
```

Protocol.md
```diff
@@ -972,13 +976,54 @@ Metadata Length = 4MB
...
+(M)etadata present = 0
+(F)ollows = 0
+(N)ext = 1
+(C)omplete = 1
+
+0MB of METADATA
+13MB of Data
+```
+
+__Note__, `complete` flag MUST be set only in the last FRAME which does not have `follows` flag set to indicate that the whole sequence is a Payload with `next` and `complete` flags at the same time.
```
rstoyanchev, line 1023:
```suggestion
__Note__, the `complete` flag MUST be set only in the last FRAME which does not have the `follows` flag set in order to indicate that the entire sequence is a `PAYLOAD` with `next` and `complete` both set.
```

Protocol.md
```diff
@@ -987,6 +1032,8 @@ When fragmented, the Metadata MUST be transmitted completely before the Data.
 
 For example, a single PAYLOAD with 20MB of Metadata and 25MB of Data that is fragmented into 3 frames:
 
+##### Request Response
```
rstoyanchev, line 1035: Similar suggestion here, to remove the sub-headings.

Protocol.md
```diff
@@ -1001,6 +1048,44 @@ Metadata Length = 16MB
...
+(F)ollows = 0
+(N)ext = 1
+
+0MB of METADATA
+13MB of Data
+```
+
+If the Requester wants to cancel sending a fragmented sequence, it MAY send a CANCEL frame without finishing delivery of the fragments. 
+
+##### Request Channel with `complete` flag
```
rstoyanchev, line 1069:
```suggestion
Below is an example for Request Channel with the `complete` flag indicating the entire sequence is a `PAYLOAD` with `next` and `complete` both set.
```

Protocol.md
```diff
@@ -1010,11 +1095,15 @@ Metadata Length = 4MB
...
 (M)etadata present = 0
 (F)ollows = 0
+(N)ext = 1
+(C)omplete = 1
 
 0MB of METADATA
 13MB of Data
 ```
 
+__Note__, `complete` flag MUST be set only in the last FRAME which does not have `follows` flag set to indicate that the whole sequence is a Request Channel with `complete` flag.
```
rstoyanchev, line 1105:
```suggestion
__Note__, the `complete` flag MUST be set only in the last FRAME which does not have the `follows` flag set in order to indicate that the entire sequence is a `PAYLOAD` with `next` and `complete` both set.
```

rstoyanchev (changes requested):

Protocol.md
```diff
@@ -949,11 +949,14 @@ When a PAYLOAD is fragmented, the Metadata MUST be transmitted completely before
 
 For example, a single PAYLOAD with 20MB of Metadata and 25MB of Data that is fragmented into 3 frames:
 
+##### Payload with `next` flag set
+
 ```
```
OlegDokuka, line 954: move to a separate doc

PR #325, opened 2022-03-29
Update specification with supporting of unbounded mode for request(n)

ghost:
Update specification with supporting of unbounded mode for request(n) to overcome absence of Magic Number from Reactive Streams specification

### Motivation:
In Reactive Streams specification, and in Project Reactor (implementation of spec), there is possibility to make stream `unbounded` by using magic number (2^63-1 (java.lang.Long.MAX_VALUE)) when calling request(n) on subscription.
But in RSocket there is no this magic number, as it uses 4 bytes instead of 8.
So `rsocket-java`, emulates `unbounded` mode by sending Int.MAX_VALUE instead or magic number if it appears.
And also current implementation can go into an `unbounded` mode when during stream lifetime, SUM of all request(n) for stream is more than Int.MAX_VALUE. After stream becomes `unbounded`, there will be no more REQUEST_N frames sent there.

The problem is, that, if implementation don't do the same thing (because this behavior isn't specified in specification), those Int.MAX_VALUE credits can be exhausted in relatively small period of time in real use case: 12-15 hours, as found by @marcelja.

### Modifications:

Adding `U` bit to REQUEST_N, REQUEST_STREAM and REQUEST_CHANNEL frames to enable `unbounded` mode and make it possible to better interop with Reactive Streams specification.

### Result:

Implementation SHOULD handle such bit, and make it possible to request `unbounded` stream.
F.e. in `rsocket-java` there should be mapping when using `request(n)` where n is of Long type (so magic number is available) to use unbounded mode if Long.MAX_VALUE appears.

### DIFF
--- a/Protocol.md
+++ b/Protocol.md
@@ -532,7 +532,7 @@ Frame Contents
     +-----------+-+-+-+-------------+-------------------------------+
     |Frame Type |0|M|F|    Flags    |
     +-------------------------------+-------------------------------+
-    |0|                    Initial Request N                        |
+    |U|                    Initial Request N                        |
     +---------------------------------------------------------------+
                           Metadata & Request Data
 ```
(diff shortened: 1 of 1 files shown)


viglucci:
Have we explored alternatively adding a new Flag to appropriate frames to denote Unbounded, rather than repurposing the proposed bit? I would be concerned that the proposed could be troublesome for implementations that are currently working with the existing understanding, and could potentially introduce interop issues for implementations lagging behind the spec if the change were to land as proposed.

Adding a new Flag feels more "additive" and non-breaking than repurposing this bit as currently proposed.

```
    |                           Stream ID                           |
    +-----------+-+-+-+-------------+-------------------------------+
    |Frame Type |0|M|F|U|    Flags    |
    +-------------------------------+-------------------------------+
    |0|                    Initial Request N                        |
                          Metadata & Request Data
```
^-- `U` flag to denote unbounded demand. Initial Request N and all REQUEST_N frames can be ignored.

yuriykulikov:
It seems that adding an U flag will be a breaking change, because when an U-flag-enabled client communicates with an outdated server, the server will be unaware of the unbounded mode and won't send any messages. Using the U-bit, however, will be considered as a large credit by the server and communication will continue to work until the credit is exhausted.

PR #327, opened 2022-07-05
graphql over rsocket spec

OlegDokuka:
Adds a separate dedicated extension spec that describes integration with GraphQL protocol

### Motivation:

GraphQL is a popular API protocol that does not has not direct binding to a specific transport. Thus, it is going to be valuable to add an addition that clearly explains how such protocol should be handled when it is run over RSocket

### DIFF
--- a/Extensions/GraphQL.md
+++ b/Extensions/GraphQL.md
@@ -1,0 +1,37 @@
+# GraphQL Extension Spec
+
+_This extension specification is currently incubating.  While incubating the version is 0._
+
+## Introduction
+
+GraphQL protocol is one of the popular API protocol for building modern edge-network communication. This extension specification provides an interoperable structure for metadadata payloads and sub-specification which defines graphql communication over RSocket protocol.
+
+## Handling GraphQL operation
+
+A GraphQL operation can be handled if it is agreed during the RSocket `SETUP` phase and specified in the data-mime type `application/graphql` or if composite metadata enabled, specified in the REQUEST PAYLOAD data mime-type `application/graphql` via [Stream Data MIME Types Metadata Extension](/PerStreamDataMimeTypesDefinition.md).
+
+GraphQL operation can be requested via `REQUEST_RESPONSE` and `REQUEST_STREAM` requests of RSocket Protocol.
+
+> Note: +type semantic may specify graphQL message encoding if needed (e.g. `application/graphql+cbor`). By default it is JSON encoding.
+
+### Single result operation
+
+To handle `query` or `mutation` GraphQL operation a `REQUESTER` MUST issue a `REQUEST_RESPONSE` payload.
+
+GraphQL operation lifecycle is bound to the RSocket REQUEST RESPONSE operation semantic and must follow it as specified by RSocket protocol.
+
+### Streaming operation
+To handle `subscription` GraphQL operation a `REQUESTER` MUST issue a `REQUEST_STREAM` payload.
+
+Streaming operation lifecycle is bound to the RSocket REQUEST STREAM operation semantic and must follow it as specified by RSocket protocol.
+
+### Handling Graphql errors
+
+GraphQL errors must be encoded as a normal RSocket `PAYLOAD` with `NEXT` semantic followed by `PAYLOAD` with `COMPLETE` flag (onNext(ERROR_MESSAGE) -> onComplete()). RSocket Protocol `ERROR` must be treated as unexpected interruption of the subscription and MUST not contain any GraphQL data in the payload.
+
+
+### Handling unexpected
+
+1. If `REQUEST_STREAM` request does not contain `subscription` GraphQL operation, such request must be rejected with an `ERROR` having `INVALID` as an error code.
+2. If `REQUEST_RESPONSE` request does not contain `query` or `mutation` GraphQL operation, such request must be rejected with an `ERROR` having `INVALID` as an error code.
+3. Hand


linux-china (approved):

Extensions/GraphQL.md
```diff
@@ -0,0 +1,37 @@
...
+### Handling Graphql errors
+
+GraphQL errors must be encoded as a normal RSocket `PAYLOAD` with `NEXT` semantic followed by `PAYLOAD` with `COMPLETE` flag (onNext(ERROR_MESSAGE) -> onComplete()). RSocket Protocol `ERROR` must be treated as unexpected interruption of the subscription and MUST not contain any GraphQL data in the payload.
+
+
+### Handling unexpected
+
+1. If `REQUEST_STREAM` request does not contain `subscription` GraphQL operation, such request must be rejected with an `ERROR` having `INVALID` as an error code.
+2. If `REQUEST_RESPONSE` request does not contain `query` or `mutation` GraphQL operation, such request must be rejected with an `ERROR` having `INVALID` as an error code.
+3. Hand
```
spencergibb, line 37: Hanging bullet item

spencergibb (approved):

Extensions/GraphQL.md
```diff
@@ -0,0 +1,37 @@
+# GraphQL Extension Spec
+
+_This extension specification is currently incubating.  While incubating the version is 0._
+
+## Introduction
+
+GraphQL protocol is one of the popular API protocol for building modern edge-network communication. This extension specification provides an interoperable structure for metadadata payloads and sub-specification which defines graphql communication over RSocket protocol.
```
yschimke, line 7: nit

```suggestion
GraphQL protocol is one of the popular API protocols for building modern edge-network communication. This extension specification provides an interoperable structure for metadadata payloads and sub-specification which defines graphql communication over RSocket protocol.
```

Extensions/GraphQL.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Handling GraphQL operation
+
+A GraphQL operation can be handled if it is agreed during the RSocket `SETUP` phase and specified in the data-mime type `application/graphql` or if composite metadata enabled, specified in the REQUEST PAYLOAD data mime-type `application/graphql` via [Stream Data MIME Types Metadata Extension](/PerStreamDataMimeTypesDefinition.md).
```
yschimke, line 11: Is this inline with https://graphql.org/learn/serving-over-http/, which defines that mime type? Should we reference that?

Extensions/GraphQL.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Single result operation
+
+To handle `query` or `mutation` GraphQL operation a `REQUESTER` MUST issue a `REQUEST_RESPONSE` payload.
+
+GraphQL operation lifecycle is bound to the RSocket REQUEST RESPONSE operation semantic and must follow it as specified by RSocket protocol.
+
+### Streaming operation
+To handle `subscription` GraphQL operation a `REQUESTER` MUST issue a `REQUEST_STREAM` payload.
+
```
yschimke, line 25: I guess it never got standardised but do any potential backends support live queries?

https://github.com/graphql/graphql-spec/issues/386

https://medium.com/open-graphql/graphql-subscriptions-vs-live-queries-e38302c7ab8e

Mainly because this is where facebooks use of rsocket came from.

Extensions/GraphQL.md
```diff
@@ -0,0 +1,37 @@
...
+Streaming operation lifecycle is bound to the RSocket REQUEST STREAM operation semantic and must follow it as specified by RSocket protocol.
+
+### Handling Graphql errors
+
+GraphQL errors must be encoded as a normal RSocket `PAYLOAD` with `NEXT` semantic followed by `PAYLOAD` with `COMPLETE` flag (onNext(ERROR_MESSAGE) -> onComplete()). RSocket Protocol `ERROR` must be treated as unexpected interruption of the subscription and MUST not contain any GraphQL data in the payload.
```
yschimke, line 30: This doesn't seem right to me. 

Compare to https://graphql.org/learn/serving-over-http/#response and https://spec.graphql.org/June2018/#sec-Errors

The spec does say "Event streams may complete in response to an error or simply because no more events will occur." so I'm happy if you know it's correct.

Isn't it valid that a stream has different errors per response, and continues?

Or is this a more serious error like a 501 in HTTP? Maybe this just needs some distinction like request level errors?

yschimke (approved):
LGTM. Seems appropriate for an incubating spec.

Extensions/GraphQL.md
nkristek, line 11: Nitpick:
```suggestion
A GraphQL operation can be handled, if it is agreed upon during the RSocket `SETUP` phase by specifying the data mime-type as `application/graphql` or if composite metadata is enabled and the data mime-type in the REQUEST PAYLOAD is specified as `application/graphql` via [Stream Data MIME Types Metadata Extension](/PerStreamDataMimeTypesDefinition.md).
```

nkristek (approved):

Extensions/GraphQL.md
rstoyanchev: There is a distinction between a "request error" and a "partial response".

A request error happens before a GraphQL request is accepted for execution by the GraphQL engine (e.g. parse error, query syntax error, security error, etc.) in which case the whole request fails. I think this should result in an ERROR frame.

Once the GraphQL engine starts execution, the response is considered valid but potentially partial. That is, there may be failures for individual fields (e.g. couldn't fetch the user's LinkedIn profile) which is reflected with a null value + a field specific error in the response. This is called a partial response. A subscription may have such partial responses but those do not mean the subscription ends.

Extensions/GraphQL.md
```diff
@@ -0,0 +1,37 @@
...
+
+
+### Handling unexpected
+
+1. If `REQUEST_STREAM` request does not contain `subscription` GraphQL operation, such request must be rejected with an `ERROR` having `INVALID` as an error code.
+2. If `REQUEST_RESPONSE` request does not contain `query` or `mutation` GraphQL operation, such request must be rejected with an `ERROR` having `INVALID` as an error code.
```
rstoyanchev, line 36: These are not really possible to validate upfront. When a GraphQL request is received, it is not yet parsed. Once it's handed to the GraphQL engine, it does the parsing and decided how to process. 

Once the GraphQL engine produces a response, e.g. `graphql.ExecutionResult`, we can check if the "data" in that is a `Publisher`, and only then can we validate something like this. 

There maybe other ways to hook this, e.g. through a `graphq.execution.Instrumentation` but generally I don't know if we really need this validation.

viglucci: > There maybe other ways to hook this, e.g. through a graphql.execution.Instrumentation but generally I don't know if we really need this validation.

Is this specifically a GraphQL Java construct? If so I would hesitant to include anything in this sub-spec that is only reasonably achievable by a single Language or existing libraries for a specific language.

PR #330, opened 2022-10-24
Protocol - fragment size clarification

jell-o-fishi:
Java implementation regards the user specified fragment size property as the size of the entire frame containing the fragment.
Suggest adding this to spec as recommendation for consistency across implementations.
Is this the right place for this ?


jell-o-fishi:
bad commit.

